PR #110, opened 2013-11-09
Allow parse content using Twig (.html files).

msosvi:


PhrozenByte:
I was recently thinking about this idea while working on Pico 2.0 (#334) and thinking about Pico 3.0 (#317), but came to the decision, that we shouldn't explicitly allow replacing Markdown with other "content processors" (like Twig).

If you want to use Twig for some "more complex" pages you can already do that. Simply create a `content/my_complex_page.md` with

```yaml
---
Template: my_complex_page
---
```

and a matching `themes/my_theme/my_complex_page.twig`. Add your contents to the Twig template and you're ready to go.

If you want to create a larger number of such "complex" pages and don't want to create both a (basically empty) `.md` file and a `.twig` template, you can also write a plugin that searches for `.twig` files in e.g. the `themes/my_theme/complex_content/` directory, adds them to Pico's pages array (`onPagesLoaded` event) and intervenes appropriately when a "complex" page is requested (basically hook into the `onRequestUrl` event to detect the request and change the `$templateName` argument of the `onPageRendering` event). However, I don't think that this feature is widely needed, thus this isn't going to make it into Pico's core.

If you rather want to extend the Markdown processor (i.e. the `Parsedown` class) with some custom features (like #318), you can do this starting with Pico 2.0. There's a new `onParsedownRegistered` event that passes Pico's `$parsedown` object as reference, allowing you to replace it with another object (however, the object's class must extend the `Parsedown` class, otherwise you'll break Pico).

If you want to *completely* replace Markdown, YAML and/or Twig with something different, you can extend Pico's class and replace the respective methods as you want (Pico is open source). We obviously will neither officially support this nor will this work with all existing plugins or themes out there, but you definitely *can* do this if you want to.

Thank you for your feedback anyway! 👍

Issue #130, opened 2014-01-02
Code block doesn't seem to handle comments in code


berthelemy:
Hi there,

I've got a code block defined in my content page which has comments embedded in it as:

/**
- A comment
- Another comment

*/

These aren't being rendered by Pico, except as blank lines.

Running it through the PHPMarkdown script works as expected, but not in Pico.

Any ideas?

Thanks,

Mark

Issue #67, opened 2013-08-21
Bug: pico removes ALL comments from the content


p-meier:
Problem
--------------------

I recently wrote an article with source code in scala - and pico stripped the scaladoc completely. 

This is the markdown for the article:
```
...

Here's the code from the ``PageRecommender`` trait. 

	trait PageRecommender {

	  /**
	   * Gets some urls and returns the recommended ones - based on the given data.
	   *
	   * @param urls - base data for recommendation
	   * @return recommended urls
	   */
	  def recommend(urls: Traversable[String]): Traversable[String]
	}

Here's the code from the ``DeliciousUserBasedPageRecommender`` class. It does all the heavy work of finding some new and possibly interesting URLs. 

...
```

After pico processed it, this came out:

```
...

Here's the code from the PageRecommender trait.

trait PageRecommender {



  def recommend(urls: Traversable[String]): Traversable[String]
}
Here's the code from the DeliciousUserBasedPageRecommender class. It does all the heavy work of finding some new and possibly interesting URLs.

...
```

Possible Solution
--------------------

I searched a little bit in the source-code and found this line in ``lib/pico.php``, causing the issue:

```php
...
     $content = preg_replace('#/\*.+?\*/#s', '', $content); // Remove comments and meta
...
```

I also found a possible fix, that works for me as a workaround:

```php
...
     //Added number 1 as third argument, to only replace the first finding!
     $content = preg_replace('#/\*.+?\*/#s', '', $content, 1); // Remove comments and meta
...
```

Thanks for the great work so far. Hope this helps.

PR #131, opened 2014-01-05
Don't remove all comments while parse content

Frodox:
There is already openned bug #67, #130 (pico removes ALL comments from the content).
This patch fix it, by removing only first comment-block (with meta info).

Now, pico.php during parsing remove all `/* such comment blocks */`
prons?
* you can use such comments for private comments, which will see nobody except you 

cons?
* it's undeclared ability, coz original Markdown language (i.e. php-markdown parser) hasn't syntax-abillity
 for comments like in other langs, and people hope so. 

So, if someone often uses pivate-comments, he can do something like `[c]:с (some comment)`, which is syntactically correct. But when people post `code` with `/* doc-comments */`, they face trouble with which they can not do anything.

Or, it will be good to be documented in Doc :)

### DIFF
--- a/lib/pico.php
+++ b/lib/pico.php
@@ -136,7 +136,7 @@ class Pico {
 	 */
 	protected function parse_content($content)
 	{
-		$content = preg_replace('#/\*.+?\*/#s', '', $content); // Remove comments and meta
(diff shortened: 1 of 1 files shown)


theshka:
+1

PR #150, opened 2014-02-27
Localize date meta

pschmitt:
Use strftime() instead of date() so that non-english websites can
display the date meta in their locale. The locale can be set with
`setLocale(LC_ALL, 'YOURLOCALE')`.
Please note that this changes the date format.
More information: http://php.net/strftime

### DIFF
--- a/lib/pico.php
+++ b/lib/pico.php
@@ -173,7 +173,7 @@ class Pico {
 			}
 		}
 		
-		if(isset($headers['date'])) $headers['date_formatted'] = strftime($config['date_format'], strtotime($headers['date']));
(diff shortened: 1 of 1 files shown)


dkyme:
Since I'm a php-non-expert, where do I got to add the setLocale?
My German Page looks strange with English Months

PhrozenByte:
@dkyme, simply add e.g.
```php
setlocale(LC_TIME, 'de_DE.utf8');
```
to your `config/config.php`. You can get a list of available locales by running `locale -a` on your server's shell.

The use of `strftime()` and `setlocale()` is fraught with problems in some environments, thus this is kind of a workaround; we're planning to release a official i18n plugin with Pico 1.1, see #334.

dkyme:
Thank you @PhrozenByte , works fine!

PR #169, opened 2014-05-26
HTTPS fix

auyongcheemeng:
Some web server returns a $_SERVER['HTTPS'] = 'no value' for non https connections

ie. nginx-1.7.0 with php5-fpm, causing all navigation links to link to https when on http version of site

https://stackoverflow.com/questions/1175096/how-to-find-out-if-you-are-using-https-without-serverhttps

### DIFF
--- a/lib/pico.php
+++ b/lib/pico.php
@@ -309,7 +309,7 @@ class Pico {
 	protected function get_protocol()
 	{
 		$protocol = 'http';
-		if(isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off'){
+		if(isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != ''){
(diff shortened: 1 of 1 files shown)


lib/pico.php
```diff
@@ -309,7 +309,7 @@ protected function base_url()
 	protected function get_protocol()
 	{
 		$protocol = 'http';
-		if(isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off'){
```
theshka: I agree with @auyongcheemeng, this is the best way to test for HTTP/S

```php
if (isset($_SERVER['HTTPS']) && $_SERVER['HTTPS'] != 'off') { 
     //HTTPS 
}
```

maybe throw in a `|| $_SERVER['SERVER_PORT'] == 443` to be super-sure..

PR #211, opened 2015-03-18
Improve project organization

dmelo:
It is a good practice to keep all the vendor directory out of git. The same goes to files that are meant to be edited, like config.php and the content directory.

In the case of config.php, it was moved to config.php.template, and "config.php" is already inside .gitignore.

The content directory was moved to content-sample. CONTENT_DIR constant is now replaced by $config['content_dir'] (on config.php.template) and it's default value is 'content-sample'.

Instructions about how to install and run the project are added on README.md.

I believe it improves the way the project is organized.


dmelo:
bump

derhuerst:
Have a look [how Kirby structures its git repo](https://github.com/getkirby/starterkit).

> It is a good practice to keep all the vendor directory out of git.

You could use git subrepos to link to the original vendor repo. That's a clean and practical solution.

> The same goes to files that are meant to be edited, like config.php and the content directory.

In case you split up the project into a core and a deploying repo (like Kirby did), I think its not necessary anymore.

dmelo:
Hi @derhuerst . About git subrepos, I think it could be a solution. But composer already covers that. It would be redundant to manage the dependencies both through git submodule and composer. Moreover, composer is (in my opinion) far better choice for dependency management.

Given that you have chosen composer for dependency management, you don't need to keep your dependencies (even as submodules) inside your project. Projects like [zf2](https://github.com/zendframework/zf2) and [apigility](https://github.com/zfcampus/apigility.org) are good examples of how to use composer with git.

I'm not sure if I understood you last remark. The config.php is a file that the user must edit. But you don't want those changes to be tracked by the Pico repo. A good solution is to 1) create a config.php.template file, that is tracked by git and 2) list the config.php into the .gitignore. This way, the user can copy config.php.template to config.php and edit it, without get the changes tracked on Pico.

dmelo:
bump

derhuerst:
I agree.

dmelo:
am I missing something? There are many open pull requests for this projects and the latest closed pull request is from February 2014. Is this project continuing on another fork, or something?

PR #219, opened 2015-04-30
Pages are now parsed by PHP first.

gixxy:
PHP in your markdown pages will now be run before the markdown is parsed.

### DIFF
--- a/lib/pico.php
+++ b/lib/pico.php
@@ -47,10 +47,16 @@ class Pico {
 
 		$this->run_hooks('before_load_content', array(&$file));
 		if(file_exists($file)){
-			$content = file_get_contents($file);
+			ob_start();
+			include($file);
+			$content = ob_get_contents();
+			ob_end_clean();
 		} else {
 			$this->run_hooks('before_404_load_content', array(&$file));
-			$content = file_get_contents($settings['content_dir'] .'404'. CONTENT_EXT);
+			ob_start();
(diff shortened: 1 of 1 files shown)


theshka:
Is it wise to include php in your markdown pages? Why not in a theme/template or plugin?

gixxy:
Useful for any case where you want more dynamic content on individual pages. 
IE: You want a different random quote or image to be displayed on a page each time it is loaded.
IE: You want to load data in a separate file or site (Like when making API Calls) and load it into pages.

Though thinking this through, this could add a tad bit of overhead, I will rewrite it as an option that is off by default and can be turned on in the config.

gixxy:
You can add PHP page parsing support via plugin, but to do so you basically need to bypass the calls to file_get_contents. which would add even MORE overhead to Pico.

theshka:
In the case of dynamic content on pages, that should be achieved through a plugin. If you follow MVC design Pico would look something like this; 

Model - Database - Markdown Files
View - Layout - Theme/Twig
Controller - Logic - Pico/Plugin

I don't think including PHP in your View is good practice.

gixxy:
That is a fair point. I am primarily a C developer (and sometimes a backend web developer in Node) So I don't generally have to deal with front-end views. Thank you.

theshka:
No worries, thank you for your input. I hope you continue to find ways to use and contribute to Pico :+1:

Issue #235, opened 2015-06-17
Decoupling Markdown and Twig


SimoneS93:
Have uou ever thought about decoupling Markdown and Twig from the core, providing them as plugin? This way one could use other template engines and html formatting languages. And why dont you use a proper YAML parser?

theshka:
The project has changed quite a bit since #116 / I started collaborating, would you be interested in implementing any of these suggestions and submitting a new PR?

SimoneS93:
Sure, I'll do it over the weekend.

SimoneS93:
I did it, but I don't know if it is worth. In my opinion Pico needs a complete rewrite.

marcus-at-localhorst:
Not sure if I am right here, but I would like to second the request:
I wrote a small plugin to render content files with [Jade template syntax](http://jade-lang.com/) (using this: https://github.com/Talesoft/tale-jade)
Unfortunately Jade's is choking on the YAML block and I can't use the raw content and jades internal caching mechanism.
My Plugin is a little bit of a crutch but for a small site, it works.
So I would love to see an easier way to choose the rendering engine for content files.

It could be determined by the contents file extension

```
blog-2016-03-14.md => Parsedown
plain.html => No processing
complex. jade => jade renderer
blog-2016-04-05.textile => textile renderer
content.wiki => wikimarkup 
```

This would make it easy to choose the right markup for whatever the content is.

PhrozenByte:
As far as I understand this issue right, it is about replacing Markdown, YAML and/or Twig with something different. There are no plannings to *completely* replace these core parts of Pico, neither in the near, nor the distant future (i.e. Pico 3.0 and later, #317).

However, Pico already allows one to achieve most things people are trying to do (like using pure Twig instead of Markdown to represent contents; see https://github.com/picocms/Pico/pull/110#issuecomment-301251348), but we definitely won't support things which are against Pico's basic idea (Markdown + YAML + Twig + 🎉 = Pico - in the simplest, but still most powerful way possible). Adding more and more abstraction contradicts Pico's demand for simplicity (what also applies to the source code).

Please refer to https://github.com/picocms/Pico/pull/110#issuecomment-301251348 for a more detailed answer why this isn't going to happen, and some explanations on how to still achieve things like this.

Thank you for your feedback anyway! 👍

PR #240, opened 2015-06-27
Decouple Markdown and Twig

SimoneS93:
Trying to decouple Markdown and Twig from core.
Removes some hooks so it is no-retrocompatible.


PhrozenByte:
Cannot be merged anymore and a plugin-based solution is preferred. Please refer to #235 for the ongoing discussion about this idea.

PR #244, opened 2015-07-14
Use scandir instead of opendir to ensure alphabetical order.

muja:
`opendir` doesn't ensure returning the directory's entries in alphabetical order for `readdir` which causes some plugin dependencies to not work - many plugins (e.g. `nav_sort`) have a dependency on `adv_meta`, for instance. This pull request replaces it with `scandir` which does exactly that and ensures the plugin dependencies work on every system.

Here is a stackoverflow:
http://stackoverflow.com/a/541539/1527244

### DIFF
--- a/lib/pico.php
+++ b/lib/pico.php
@@ -359,8 +359,8 @@ class Pico
     protected function get_files($directory, $ext = '')
     {
         $array_items = array();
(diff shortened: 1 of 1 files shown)


MuhPirat:
:+1:

Issue #231, opened 2015-06-04
Crop Images plugin


microcipcip:
Is there any plugin to crop an image from a markdown page?

theshka:
Could you be more specific? Do you need to crop it dynamically? Are you just trying to resize it? Have you tried accomplishing it with CSS?

microcipcip:
Yes I mean dynamically, for example:
```
![alt text](http://www.img.jpg?width=350&height=350&crop "Title text here")
```

theshka:
I'm not aware of any Pico plugin that exists for this purpose. You could, however, consider using another class to do this. I've used this one in the past: https://github.com/claviska/SimpleImage

microcipcip:
I was thinking about something like https://github.com/mosbth/cimage or https://github.com/lencioni/SLIR , but I can't get it to work (mostly because of issues with the relative path)

Issue #223, opened 2015-05-11
PICO v.9


johnnywon:
This update is broken. 

It will not run, you require /vendor and it's not even included in the repository. Please update immediately, for all other users please use v.8 which works normally.

theshka:
Please see issues #121, #211, #216. You need to run the composer install. Instructions are available in the README.

johnnywon:
I don't see why we have to run composer install when the previous update worked without it. You've created an extra and unnecessary step.

theshka:
Because it is good practice to keep the vendor directory out of git. 
This update was committed before I started collaborating on the project.

johnnywon:
If it is "good practice of keeping the vendor director out of git" and "Pico is a stupidly simple, blazing fast, flat file CMS." then why would you intentionally make this harder? This means I can't ask my clients to just download the zip and run it on their web server, they need SSH and install something they're not familiar with. This is a significant problem to real people.

dav-m85:
I suggest providing an already "composer installed" package build for those use cases.

@theshka :+1: Like this change, vendor was a pain.
@johnnywon Sme reading on why it is a bad idea of having vendor tied to your local subversion tree: https://getcomposer.org/doc/faqs/should-i-commit-the-dependencies-in-my-vendor-directory.md

Issue #221, opened 2015-04-30
Pico v0.9 and beyond....


theshka:
I have recently started collaborating on Pico, and an issue was raised regarding the closing of stale threads and pull requests. In the interest of cleaning up, I have quickly reviewed all issues and closed any that I thought were irrelevant, resolved or abandoned. Likewise, I have merged many pull requests and begun labelling issues for discussion and better organization. 

If you think a thread was erroneously closed, please let me know. I will gladly reopen any issue or pull request anyone raises issue with. Going forward, Issues & PR's will ideally be labelled with a `type:`, `status:`, and `priority:`. 

I hope this refresh of the Pico repo will stimulate new dialog and contributions. Please feel free to discuss your thoughts below. 

Thanks, everyone! :)  :+1: 

**Note:** If anyone is interested, during Pico's decline it branched out as BaunCMS and PhilleCMS as well.. See #199, #208

johnnywon:
Respectfully, .9 needs to be fixed before this project goes forward. If "Pico is a stupidly simple, blazing fast, flat file CMS" then I don't see why the functionality of .8 was scrapped for a command line install as a pre-requisite for making Pico work. 

The dead-simple drag and drop and it works is what attracted me and my friends to Pico, v.9 absolutely should be even simpler to install and use.

theshka:
It was suggested in the issues that I linked you to earlier that; Pico be split in to two branches. One for development (no vendor) and one for production (complied). As this project was dead in the water before I arrived, I am willing to entertain this. Would you like to make the desired changes and submit a pull request for feedback?

johnnywon:
Honestly, I do not have the skills to make the desired changes. Your efforts are much appreciated, especially from people like me who are more design than PHP expert. My suggestion would be to go back to the original ethos of Pico "stupidly simple, blazing fast, flat file CMS" and keep-on with a singular production branch that might error on the side of messy but "stupidly simple" and "blazing fast." 

If it is poor practice to have a /vendor folder, it should be weighed against the simplicity of install, especially for laypersons, since it would mean greater overall usage vs. organizational efficiency.  I think the growth and simplicity of Pico is credit to it's dead simple install and we must continue the cause.

theshka:
Perhaps a composer auto-installer script could be crafted for those without SSH access. 
@johnnywon you may also want to consider using BaunCMS https://github.com/BaunCMS/Baun

johnnywon:
[...]

Issue #201, opened 2015-01-23
Nested pages variable


DoctorBlue:
It'd be nice to have the pages variable but with nesting. That way I could create submenus without too much hassle.

    {% for page in pages %}
    <li><a href="{{ page.url }}">{{ page.title }}</a></li>
        <ul>{% for subpage in page.subpages %}
         <li><a href="{{ subpage.url }}">

If you catch my drift.

PR #252, opened 2015-08-28
Pico 1.0

### LINKED ISSUE #93: [question] why do you call parse_content for each page in get_pages?
In get_pages for each page the function parse_content will be called, in my little test environment with some hooks, the method get_pages is called 5 times which calls parse_content 5 times for all 20 pages, so with every request parse_content will be called 100 times. 
Can we cache the calls (Singleton) or better, call parse_content only, if the parsed_content property is used.

### LINKED ISSUE #99: Content filenames containing spaces not working
I replaced the following line in the constructor of class Pico

     // Get our url path and trim the / of the left and the right
     :
     $url = preg_replace('/\?.*/', '', $url); // Strip query string

with:

     $url = preg_replace('/\?.*/', '', urldecode($url)); // Strip query string

Now the system is able to navigate to that files. Otherwise you always get a 404 error.

kg mcbSolutions

### LINKED ISSUE #140: Titles used as page identifiers
I couldn't figure out why {{ current_page.url }} showed the url of some other page on my front page. I had a dig around and I found these lines in Pico.php (from 74)
```php
		while($current_page = current($pages)){
			if((isset($meta['title'])) && ($meta['title'] == $current_page['title'])){
				break;
			}
			next($pages);
		}
```

This tells me that the page that is chosen as the current page is the first with the same title, and if the title isn't set, it ends up as the last page.

This kind of sucks. The docs don't say that titles are unique identifiers, and they really shouldn't be. Having two pages with the same title shouldn't break stuff. You can tell them apart by the file name.

### LINKED ISSUE #158: Performance on large number of pages
That's not really an issue.

If you are interested, I've been playing with Pico regarding to issue #97 and I made a few changes and a little performance check which gave me nice results. A brief summary in the repo https://github.com/geedmo/Pico

When you get chance, I would like to know your thoughts.

### LINKED ISSUE #171: Is there a way to make one plugin require another?
Or am I stuck with writing it all in one plugin?

### LINKED ISSUE #210: Get content file filename
Hi,

I was wondering if there was a way to get the filename of a "blog post" ?

In fact, I'd like to get the page.url and to remove the filename from it to make a link to the upper level.

Let's says I have a blog post here : http://example.com/tuto/test
I'd like to get the link: http://example.com/tuto/

Is there a way this using some kind of page.filename or using a Twig filter ?

Thanks in advance.

### LINKED ISSUE #248: First time clone and localhost Pico: Bunches of warnings
it is bad for first experience

( ! ) Warning: strtotime(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in D:\server\UwAmp\www\pico\Pico\lib\pico.php on line 189
Call Stack
#	Time	Memory	Function	Location
1	0.0000	139528	{main}( )	..\index.php:0
2	0.0020	267288	Pico->__construct( )	..\index.php:15
3	0.0030	284784	Pico->read_file_meta( )	..\pico.php:69
4	0.0030	285832	strtotime ( )	..\pico.php:189

### LINKED ISSUE #249: Allow .twig as file extension for templates
I use PhpStorm and the Twig plugin expects the file extension to be *twig* instead of *html* in order to enable Code Highlighting and Live Templates. I imagine this could be true for other IDEs too – haven't checked though. Anyways, I implemented a quick fix for myself by replacing

```php
$output = $twig->render($template . '.html', $twig_vars);
```

with this:

```php
if(file_exists(THEMES_DIR . $settings['theme'] . '/' . $template . '.twig')){
    $output = $twig->render($template . '.twig', $twig_vars);
} else {
    $output = $twig->render($template . '.html', $twig_vars);
}
```

### LINKED ISSUE #250: Incorrect versioning in file pico.php and config.php.template
I stumbled across a *minor* issue that irritated me a little. In the current release the file ```pico.php``` contains an older version number – **0.8**. Also, the file ```config.php.template``` has the version **0.9** although the latest release was **0.9.1**. Really no big deal, I was just confused for a second if I had the latest version since I downloaded the zip from (picocms.org)[http://picocms.org/] – that was before I saw that the download link referred to the GitHub master branch.

### LINKED ISSUE #251: Issues with URLs and loading files
Hi there!

Just started playing with Pico and it is possible that my error is something really stupid but since I tried a lot of things and nothing is working I'm asking you for help.

The problem is that all pages except index are showing 
```The requested URL /services was not found on this server.```
and in console
```
File does not exist: /home/ubuntu/workspace/projects/, referer: https://symbolic-benjam7n.c9.io/?_c9_id=livepreview0&_c9_host=https://ide.c9.io
```

### LINKED ISSUE #253: Problem with HTTPS
Hi,

[My site](https://github.com/botwiki/botwiki.org) wasn't working correctly with HTTPS. It looked like it was just ```get_protocol``` not returning the correct value, but it turned out [this](http://stackoverflow.com/questions/10951325/how-can-serverhttps-be-empty-when-accessing-over-https) was the problem.

I am not really sure if this was just my particular server setup, but I'm just letting you know about this.

### LINKED ISSUE #257: PHP Built-in web server start command
Then I use 
```bash
php -S 0.0.0.0:8080 ./
````
And after I try to access http://localhost:8080/ and receive 500 status.
I get in my console:
```bash
[Wed Sep 30 00:23:06 2015] PHP Warning:  Unknown: failed to open stream: Inappropriate ioctl for device in Unknown on line 0
[Wed Sep 30 00:23:06 2015] PHP Fatal error:  Unknown: Failed opening required './' (include_path='.:/usr/share/php:/usr/share/pear') in Unknown on line 0
```
But if use 
```bash
php -S 0.0.0.0:8080 -t ./
```
everything works fine.
I'm using Ubuntu 14.04 and php 5.5.9-1ubuntu4.12 (cli)

### LINKED ISSUE #267: Mod rewrite rules ignored
I'm trying to add some lines to my `.htaccess` to implement friendlier URLs, but rules I add on top of the provided `.htaccess` have no effect.

I'm sure my rules are correct, because I can observe the rewrite with the `$_SERVER['REDIRECT_URL']` php variable.

I browsed the source code and here is what I noticed: the file that pico loads is derived entirely from `$_SERVER['REQUEST_URI']` (found on [these lines](https://goo.gl/ZNp2XP)), and I'm sure that using this variable instead of `$_SERVER['REDIRECT_URL']` is what is ignoring my rewrite rules.

I love how pico is small and barebones/concise, and I really want it to stay that way. Is it within pico's scope to support mod_rewrite? As I understand, the `$_SERVER['REDIRECT_URL']` variable is an apache-specific variable, so that's another thing to consider: is pico meant to be portable across all web-servers? Or just on apache? If it's designed for portability, I guess I'll need to write a plugin to do this.

PhrozenByte:
**First of all: Why?** The sourcecode of existing forks like [BaunCMS](http://bauncms.com/) and [PhilleCMS](http://philecms.com/) isn't "stupidly simple" anymore... You can read Picos sourcecode from top to down and even copy&paste programmers will understand what's going on. It's all about "understanding at first glance".

Most important second: **All changes are 100% backward compatible.**

I considered writing something own, but then caught up with Pico. The only thing missing was clean code - Picos concept, workflow and code really is "stupidly simple", but very powerful. **Actually I just did a code refactoring.** [I don't want to fork Pico](https://xkcd.com/927/) and I won't do it even if you reject my changes (obviously I'll still use it myself :smile:).

**This should be v1.0 ready.** I recommend to release it as v1.0-beta, waiting some weeks to test it on a large user basis (I'll fix all appearing problems) and finally releasing it as Pico 1.0 :smile:

Please give me a hint if you'll merge this, I'll then update the homepage (`gh-pages` branch), `changelog.txt` etc. accordingly.

**What did I do?**
- Fixes #79, Closes #93, Fixes #99, Closes #103, Closes #116, Fixes #140, Fixes #158, Closes #171, Fixes #210, Closes #241, Closes #244, Closes #246, Fixes #248, Fixes #249, Fixes #250, Fixes #251, Fixes #253, Fixes #257, Fixes #267
- The code is now documented, code styling follows PSR. Pico superficially grows from 400 LoC to 1000 LoC, when removing all comments, Pico grows from 300 LoC to 500 LoC, mostly because of new public getter and helper methods (= 100 LoC) and PSR code styling. As said, there are no big functional changes, it's more a code refactoring.
- The new routing system now works out-of-the-box (even without rewriting) with any webserver using the QUERY_STRING routing method. Internal links now look like `?sub/page`, rewriting to basically remove the `?` is still possible and recommended. Contrary to Pico 0.9 every webserver should work just fine. Pico 0.9 required working URL rewriting, so if you want to use old plugins/themes/contents, a working rewrite setup may still be required. If you're not using the default `.htaccess`, you must update your rewrite rules to follow the new principles. Internal links in content files are declared with `%base_url%?sub/page`. Internal links in templates should be declared using the new `link` filter (e.g. `{{ "sub/page"|link }}`), what basically calls `Pico::getPageUrl()`. You musn't change anything if you setup rewriting (what was required in Pico 0.9...), so I assume this to be fully backward compatible :smile:
- I've implemented a whole new plugin system while maintaining full backward compatibility. See the class docs of `IPicoPlugin` for details. The new event system supports plugin dependencies as well as some new events. It was necessary to reliably distinct between old and new events, so all events were renamed. The new `PicoDeprecated` plugin is crucial for backward compatibility, it's enabled on demand. Refer to its class docs for details.
- I completely overhauled the sorting algorithms, so sorting by date now works as expected by using timestamps and index files are listed before their sub pages (so e.g. `sub/index.md` is listed before `sub/foo.md`).
- Inaccessible files (like `sub.md` while `sub/index.md` exists) are now excluded from the pages list.
- Now supporting per-directory `404.md` files (if there is a `sub/404.md`, this file is displayed in favour of the global `404.md` file, when e.g. the non-existing `sub/notexisting.md` is requested)
- The file extension of content files is now configurable.
- Heads up! #158 is fixed only when the `PicoParsePagesContent` plugin isn't enabled. It's disabled by default, but gets automatically enabled with `PicoDeprecated` as soon as an old plugin is loaded. This is necessary to maintain backward compatibility. You can still disable it manually by executing `$pico->getPlugin('PicoParsePagesContent')->setEnabled(false);` or adding `$config['PicoParsePagesContent.enabled'] = false;` to your `config.php`.
- The meta headers are now parsed by the YAML component of the Symfony project (see #116) and it isn't necessary to register new headers during the `onMetaHeaders` event anymore. The implicit availability of headers is supposed to be used by users and pure (!) theme developers *only*, so we're still instructing plugin developers to register their headers. See the commit message of 7537159 for more detail.
- Meta header variables are now accessible in content files using `%meta.*%`, so you mustn't repeat yourself. You can now put an excerpt into the `description` meta variable and output the same content at the beginning of the article using `%meta.description%` (see `index.md` for a example).
- Initialization of Pico now works completely different: rather than defining constants (which are probably conflicting with other applications...), Pico now expects its paths to be passed as parameters to the constructor. The constructor doesn't start Picos processing anymore, you now have to call the `Pico::run()` method, which returns the parsed page contents instead of directly `echo`ing them. The `PicoDeprecated` plugin defines the now deprecated constants `ROOT_DIR`, `LIB_DIR` etc., so old plugins can still use them. Those constants are defined before reading `config.php` in Picos root folder, so upgrading users usually aren't bothered with e.g. a `PHP Notice: Use of undefined constant ROOT_DIR - assumed 'ROOT_DIR'` error when using `ROOT_DIR` in their `config.php` (so: no BC break). New users don't need the constants anymore, relative paths are now always interpreted to be relative to Picos root dir, so `$config['content_dir'] = 'content';` is always sufficient (previously this was depending on the webserver config). All these changes are supposed to improve Picos interoperability with other applications and allows developers to integrate Pico in other applications, therefore I additionally added a new `Pico::setConfig()` method to even make the use of a `config.php` optional.
- The Twig cache dir doesn't exist anymore. The provided directory never was helpful, in most cases one still had to make the directory writable for the webserver, so it never worked out-of-the-box.
- All `index.html` files were removed; instead of helping users to secure their website, they create a false sense of security. This leads to situations like "I'm receiving a empty website when navigating to the content folder, thus everything must be safe!"
- I decided explicitly to NOT implement pages as objects ("stupidly simple", see above). Anyway, I think plugin developers shouldn't manipulate data in "wrong" events, this could lead us to unexpected behaviour. Sure, plugin developers still can do this, we're passing variables by reference, but it's not that obvious. I even thought about dereferencing the values after the corrosponding event was called, but that would be backward incompatible. What do you think?
- How to fix the "composer problem" discussed in #221 and #223? There's a very simple solution: When creating a release on GitHub (after you've pushed the tag) you can upload "binaries". Simply execute composer locally, create a ZIP archive and upload the result as "binary".
- I didn't care much about #110, #238, #239 and #240 because I simply don't need these features. But I think they are good ideas and the core should support this. Just my 2 cents :smile:
- #201 and #231 should be closed - this can easily be achieved with plugins. In fact, there are already plugins adding support for these features...
- Imo distinct documentations for users, theme designers and plugin devs is MUCH more important than unit tests... Pico is a project with just about 500 LoC (+ comments), such a manageable project doesn't necessarily require unit tests - they are nice to have, but that's it. Documentation should be the top priority!
- Note: I'm no english native speaker. Maybe someone should look through my code comments :smile:


theshka:
Great work! I would definitely like to push this upstream-- it's so fresh, so clean :+1: 

Just a couple small problems to fix straight away: 

In `global.php` there's some funky paths still set to your environment. Replacing the first two lines, and THEMES_DIR path does the trick:
```php 
//define('HTTPDOCS', realpath(rtrim(__DIR__, '/')) . '/');
//define('ROOT_DIR', realpath(HTTPDOCS . '../httpdocs-includes') . '/');
define('ROOT_DIR', realpath(rtrim(__DIR__, '/')) . '/');

//define('THEMES_DIR', HTTPDOCS . 'themes/');
define('THEMES_DIR', ROOT_DIR . 'themes/');
```

and line 377 of `pico.php` throws a NOTICE when QUERY_STRING is null. 
```php 
//$pathComponent = $_SERVER['QUERY_STRING'];
$pathComponent = (isset($_SERVER['QUERY_STRING']) ? $_SERVER['QUERY_STRING'] : '');
```

I'd like to get some more input, anyone else still following Pico??

PhrozenByte:
Oops, that was the wrong `global.php` :smile:
Sure, it's a good idea to wait for about 1-2 weeks to get a little more feedback before merging this, it's a quite big update

fourtonfish:
When ordering posts by date, would it make more sense to use something like this?

```
  if ($order_by == 'date' && isset($page_meta['date'])) {
      $sorted_pages[$page_meta['date_formatted'] . $date_id] = $data;
      $date_id++;
  } else {
```

Currently it's this.

```
      $sorted_pages[$page_meta['date'] . $date_id] = $data;
```

This seems to be the only thing that works for me, as I'm using the date format of ```MMMM DD, YYYY``` (eg August 31, 2015).

PhrozenByte:
I'm not really sure what you mean. In this PR posts aren't sorted by the `Date` meta header directly, instead they are sorted by `$data['time']`, the timestamp resulting from a `strtotime()` call. What sorting would you expect?

fourtonfish:
@PhrozenByte 

Well, if you go to https://botwiki.org/tag/twitterbot, the posts are now correctly ordered while using the ```MMMM DD, YYYY``` format (see [this](https://raw.githubusercontent.com/botwiki/botwiki.org/master/content/bots/twitterbots/holidaybot4000.md) for an example).

If I go back to using ```$page_meta['date'] ``` for sorting, it seems almost random.

Feel free to clone https://github.com/botwiki/botwiki.org and play with it locally.

PontusHorn:
This looks great to me. Would be more than happy to see it merged.

@fourtonfish, have you looked through the changes in the PR? The code you're referring to isn't used anymore in it. Your problem, as I understand it, is already fixed here.

fourtonfish:
@PontusHorn Ah, my bad, I think I only looked at the most current stable, not this PR.

Well, great job, then :-)

bricebou:
Hi,

First of all, I'm not a developer and english is not my native language ; but it's great to see Pico under active development !

I don't understand all the implications of this pull request and I can't evaluate your code, but, based on some comments, it seems great. And it's coherent with the purpose of Pico : simple. And I agree with a proposition that have been already made : Pico should be available as an archive to keep its installation simple and the more wide as possible (you don't always have an SSH access to perform composer on shared hosting).

I've written some small plugins for Pico and I was wondering what would be necessary to make them coherent with the changes you made. I'd like them to be fully compatible with this new version, and not use the deprecated "way".

Would it be possible for someone here to review my code and give me some advice, recommendations... to perform an update. 
Here are the repositories:
- https://github.com/bricebou/pico_leaflet (adding Leaflet to Pico) 
- https://github.com/bricebou/pico_SyntaxHighlighter (SyntaxHighlighter for Pico)
- https://github.com/bricebou/pico_piwik (Piwik for Pico)
- https://github.com/bricebou/pico_ganalytics (Google Analytics)

Thanks in advance and thanks for all the work @PhrozenByte, @theshka and others made !

PhrozenByte:
@bricebou, sure, that's pretty easy. The most important changes are regarding the event system. You can use the transition table at the `PicoDeprecated` [class docs](https://github.com/PhrozenByte/Pico/blob/pico1.0/plugins/00-PicoDeprecated.php) to see the new event names. Some events changed their parameters to be more consistent, so you should compare them to the [sourcecode](https://github.com/PhrozenByte/Pico/blob/pico1.0/plugins/DummyPlugin.php) of `DummyPlugin` (formerly `Pico_Plugin`) - or simply use the list below. Then add `extends AbstractPicoPlugin` to your class and you're done :smiley:

If you're using internal links to other pages, you shouldn't use `$settings['base_url']` anymore. Use `$this->getPageUrl($page)` instead, otherwise your plugin requires a working URL rewriting setup. I recommend to delete the `.htaccess` file while developing, that's the easiest way to guarantee support of the new routing system.

It's recommended, but not required, that you rename your classes to follow the PSR principles, so e.g. `Pico_Leaflet` becomes `PicoLeaflet` and the plugin file name becomes `PicoLeaflet.php`.

Here's a list of changes to the event system that are relevant for existing plugins:
* All events were renamed - see the transition table at the `PicoDeprecated` [class docs](https://github.com/PhrozenByte/Pico/blob/pico1.0/plugins/00-PicoDeprecated.php) for details
* You can access all Pico variables using public getter methods, so instead of `public function request_url(&$url) { $this->url = $url; }` to remember the requested URL you can simply use `$this->getRequestUrl()`
* `onContentLoaded` and `on404ContentLoaded` dropped `$requestFile` - use `$this->getRequestFile()` instead
* `onSinglePageLoaded` dropped `$meta` - use `$pageData['meta']` instead
* `onSinglePageLoaded` passes `$pageData` indexed by ID - use `array_values($pageData)` when necessary
* `onPageRendering` changed its parameter order to `$twig, $twigVariables, $templateName`
* `onPageRendering` now passes `$templateName` with its file extension

bricebou:
Thanks for these precisions. I'll look into it, it will be my second semester's hobby...

If I understand correctly, each plugin should have its own config/config.php file ?

PhrozenByte:
No, I didn't change anything regarding configuration, it's still all inside a single `config/config.php`.

@all 
There are some "What do you think?" questions in the PR message, the most important about the need to register new meta header variables during `onMetaHeaders`. It would be nice to get a little feedback about that, I'm currently tending to remove the need to register new meta variables (as it is required in Pico 0.9). Registering a meta variable still guarantees the existance of the array key, so you don't need `isset()` checks. It also still allows plugin developers to access a meta variable through another array key.

bricebou:
Ah ok, I've misunderstood lines 36 to 38 in https://github.com/PhrozenByte/Pico/blob/pico1.0/plugins/00-PicoDeprecated.php

Thanks again.

PontusHorn:
@PhrozenByte I think it's reasonable to require meta headers to be registered, as it makes the usage of the header more predictable. Not a huge issue either way though.

lib/Pico.php
```diff
@@ -0,0 +1,943 @@
...
+        $this->triggerEvent('onPageRendered', array(&$output));
+
+        echo $output;
+    }
+
+    /**
+     * Loads plugins from PLUGINS_DIR in alphabetical order
+     *
+     * Plugin files may be prefixed by a number (e.g. 00-PicoDeprecated.php)
+     * to indicate their processsing order. You MUST NOT use prefixes between
```
PontusHorn: Typo in "processing".

lib/Pico.php
```diff
@@ -0,0 +1,943 @@
...
+     * Parses the file meta from raw file contents
+     *
+     * Meta data MUST start on the first line of the file, either opened and
+     * closed by --- or C-style block comments (deprecated). The headers are
+     * parsed by the YAML component of the Symfony project. You MUST register
+     * new headers during the `onMetaHeaders` event first, otherwise they are
+     * ignored and won't be returned.
+     *
+     * @see    <http://symfony.com/doc/current/components/yaml/introduction.html>
+     * @param  string $content the raw file contents
```
PontusHorn: Wrong parameter name - should be $rawContent

lib/Pico.php
```diff
@@ -0,0 +1,943 @@
...
+    {
+        return $this->getBaseUrl() . ((!$this->isUrlRewritingEnabled() && !empty($page)) ? '?' : '') . $page;
+    }
+
+    /**
+     * Recursively walks through a directory and returns all containing files
+     * matching the specified file extension in alphabetical order
+     *
+     * @param  string $directory start directory
+     * @param  string $ext       return files with this file extension only (optional)
```
PontusHorn: Wrong parameter name - should be $fileExtension

lib/Pico.php
```diff
@@ -0,0 +1,943 @@
...
+                // exclude hidden files/dirs starting with a .; this also excludes the special dirs . and ..
+                // exclude files ending with a ~ (vim/nano backup) or # (emacs backup)
+                if ((substr($file, 0, 1) === '.') || in_array(substr($file, -1), array('~', '#'))) {
+                    continue;
+                }
+
+                if (is_dir($directory . '/' . $file)) {
+                    // get files recursively
+                    $result = array_merge($result, $this->getFiles($directory . '/' . $file, $fileExtension));
+                } elseif (empty($fileExtension) || (substr($file, -strlen($fileExtension)) === $fileExtension)) {
```
PontusHorn: Should use the $fileExtensionLength variable defined in line 901 instead of the strlen() call.

lib/AbstractPicoPlugin.php
```diff
@@ -0,0 +1,248 @@
...
+        if (!empty($dependants)) {
+            if ($recursive) {
+                foreach ($this->getDependants() as $pluginName => $plugin) {
+                    if ($plugin->isEnabled()) {
+                        if (!$plugin->isStatusChanged()) {
+                            $plugin->setEnabled(false, true, true);
+                        } else {
+                            throw new RuntimeException(
+                                "Unable to disable plugin '" . get_called_class() . "': "
+                                + "Required by manually enabled plugin '" . $pluginName . "'"
```
PontusHorn: Wrong string concatenation operator

lib/AbstractPicoPlugin.php
```diff
@@ -0,0 +1,248 @@
...
+                            );
+                        }
+                    }
+                }
+            } else {
+                $dependantsList = 'plugin' . ((count($dependants) > 1) ? 's' : '') . ' ';
+                $dependantsList .= "'" + implode("', '", array_keys($dependants)) + "'";
```
PontusHorn: Wrong string concatenation operator

lib/AbstractPicoPlugin.php
```diff
@@ -0,0 +1,248 @@
...
+                throw new RuntimeException(
+                    "Unable to disable plugin '" . get_called_class() . "': "
+                    + "Required by " . $dependantsList
```
PontusHorn: Wrong string concatenation operator

plugins/DummyPlugin.php
```diff
@@ -0,0 +1,267 @@
...
+    public function prepareFileContent(&$content)
+    {
+        // your code
+    }
+
+    /**
+     * Triggered after Pico parsed the contents of the file to serve
+     *
+     * @see    Pico::getFileContent()
+     * @param  stirng &$content parsed contents
```
PontusHorn: Typo in "string"

PhrozenByte:
Thanks @PontusHorn 

@theshka: I suggest to wait 1-2 more weeks, I'm still hoping for more feedback (+ I currently don't have time to update the docs :smile:)

@all: Feedback is still appreciated!

composer.json
```diff
@@ -14,9 +14,14 @@
     "require": {
         "php": ">=5.3.2",
         "twig/twig": "1.18.*",
-        "erusev/parsedown-extra": "dev-master@dev"
+        "erusev/parsedown-extra": "dev-master@dev",
```
dav-m85: Mh why not a stable constraint here ? Gonna break at the next composer update.

lib/IPicoPlugin.php
```diff
@@ -0,0 +1,97 @@
...
+ * the events introduced in Pico 1.0 when releasing a new version of your
+ * plugin. If you want to use any of the new events, you MUST implement
+ * IPicoPlugin and update all other events you use.
+ *
+ * @author  Daniel Rudolf
+ * @link    http://picocms.org
+ * @license http://opensource.org/licenses/MIT
+ * @version 1.0
+ */
+interface IPicoPlugin
```
dav-m85: Been a while I haven't seen hungarian notation. I would suggest considering the sf2 usage, tht would be PicoPluginInterface

dav-m85:
Nice work @PhrozenByte. That's a LOT of work you put it here, I've been myself tempted of forking but always get back to the source. If you're sure the changes does not introduces any BC breaks, I would suggest sticking to semver and just increment the minor version.

Also I find really difficult to review all changes in one single PR, everything is mixed. It would be IMHO simpler to split the PR in multiple sub components and submit them to a temporary 0.9-dev branch. @theshka ?

I would consider switching to 1.0 branch once we got some testing coverage and definitely remove the stuff you introduce as deprecated in this PR.

Lomanic:
Hi, thank you for your work on Pico @PhrozenByte !

Unfortunately, I tried to install your pico1.0 branch with **PHP 5.3.3-7** and Apache with mod-rewrite enabled like below with the following result when testing in the browser:
```
Fatal error: Can't use method return value in write context in /var/home/XXX/www/picocms4/lib/Pico.php on line 838 
```
Steps I used to install:
```bash
cd www
wget 'https://codeload.github.com/PhrozenByte/Pico/zip/pico1.0' -O 'Pico-pico1.0.zip'
unzip Pico-pico1.0.zip
mv Pico-pico1.0 picocms4
cd picocms4
curl -sS https://getcomposer.org/installer | php
php composer.phar install
```

I just replaced this problematic line by `if ($this->getConfig('base_url')) {` following [this SO answer](http://stackoverflow.com/a/4328049) and then it complained that there were no config/config.php file to include, so I copied the `config/config.php.template` file to `config/config.php`, and now there are a lot more errors when testing in browser:
```
Notice: Undefined variable: config in /var/home/XXX/www/picocms4/config/config.php on line 61
Warning: DOMDocument::saveHTML() expects exactly 0 parameters, 1 given in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 492
Warning: DOMDocument::loadHTML(): Empty string supplied as input in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 471
Catchable fatal error: Argument 1 passed to DOMNode::removeChild() must be an instance of DOMNode, null given, called in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 496 and defined in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 472
```

Was PHP 5.3.3 support dropped or is it something else? I see in composer.json that it requires "php": ">=5.3.2", so it should be OK, am I missing some steps, some doc? Should I and could I open an issue for this pull request?

PhrozenByte:
No @Lomanic, you're not missing anything, you just found some bugs which should be fixed now, please try again :smiley:

@dav-m85: These are quiet big changes, so, in my opinion, releasing them as Pico 1.0 is justifiable. Please remember that Pico never had a stable release - all previous releases were pre-releases (`0.` versions), this actually would be the first stable release. There are no BC breaks, but the changes regarding the plugin system are enormous. I even recommend to keep the `PicoDeprecated` plugin for a very long time (e.g. until Pico 2.0). Pico already has a ecosystem that depends on these deprecated features, releasing a new, incompatible version cuts Pico off from its own ecosystem. I agree that at least auto-enabling the `PicoParsePagesContent` plugin should be removed with the next bigger release (e.g. Pico 1.1) - this plugin really has a big performance impact (as Pico 0.9 had...).

lib/IPicoPlugin.php
PhrozenByte: I personally prefer prefixing interfaces/traits with `I`/`T` towards adding `Interface`/`Trait` - but that's nothing I insist on. More opinions are appreciated.

composer.json
PhrozenByte: @theshka (https://github.com/picocms/Pico/commit/968dc183db9fe57ad94fb5e2a536e52fc0d13243), are there any reasons that oppose using e.g. `0.7.*` instead?

theshka: No, I think that was just overlooked at some point. I would say to use a stable constraint here instead of the dev branch.

lib/IPicoPlugin.php
theshka: Out of my depth with this one, and have no strong preference either way; however, PicoPluginInterface sounds more natural to me than IPicoPlugin. :speak_no_evil:

Lomanic:
Thanks @PhrozenByte for the quick fix!

I changed the two modified files (`config/config.php.template` and `lib/Pico.php`), copied `config.php.template` to `config.php` and got these errors:
```
Warning: DOMDocument::saveHTML() expects exactly 0 parameters, 1 given in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 492
Warning: DOMDocument::loadHTML(): Empty string supplied as input in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 471
Catchable fatal error: Argument 1 passed to DOMNode::removeChild() must be an instance of DOMNode, null given, called in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 496 and defined in /var/home/XXX/www/picocms4/vendor/erusev/parsedown-extra/ParsedownExtra.php on line 472
```


After some research, it looks like Parsedown-extra uses [DOMDocument::savehtml()](http://www.php.net/manual/en/domdocument.savehtml.php) with the $node parameter, and so we need PHP 5.3.**6** (I have 5.3.3 ; see savehtml()'s [Changelog](http://www.php.net/manual/en/domdocument.savehtml.php#refsect1-domdocument.savehtml-changelog)). You will have to raise the php version requirement in composer.json (Parsedown-extra should do it itself, I'm surprised it does not include a minimum php version in its own composer.json). I will perhaps downgrade Parsedown-extra until it does not use this function (this is just for testing, I will then migrate to a better updated environment). I will report which version works with php 5.3.3.

PhrozenByte:
@Lomanic: Can you please report this issue to [erusev/parsedown-extra](https://github.com/erusev/parsedown-extra)? Maybe they'll fix this problem before we merge this PR.

Lomanic:
OK, I migrated my old pico (v0.8) to your v1.0 fork install, moved my plugins, moved my theme, moved my old config.php to `config/config.php`.

To migrate from my previous pico version (v0.8 I think) I copied my old `config.php` to `config/config.php.template` and added `return $config;` at the end of the file, otherwise I just had the default configuration. I copied my theme and my plugins to the right dirs and so far my theme is not broken. Unfortunately, I don't have excerpts of my articles on the default page (as a blog), with the `{{ page.excerpt }}` twig var.

Tested plugins:
 * Lomanic/Pico-Editor-Plugin : can login but all the paths with slashes are broken
 *  rewdy/Pico-Pagination : paths with slashes are broken, but listing on the 1st page works
 * pico_search is broken for the same reason

PhrozenByte:
Excerpts were removed by default, you can re-enable them by adding `$config['PicoExcerpt.enabled'] = true;` to your `config.php`. `PicoExcerpt` depends on `PicoParsePagesContent`, what eliminates all performance advantages of Pico 1.0. You should move the excerpt contents to the `Description` meta header and add `%meta.description%` to the beginning of your content file. By the way, `PicoExcerpt` gets automatically enabled when a old plugin is loaded, thus this is no BC break.

You moved your `config.php` to `config/config.php.template`? This should disable your custom configuration and Pico falls back to its default configuration.

Can you please provide a copy of your website for debugging? Simply leave your contents directory out. Please send me an email through the contact form at http://daniel-rudolf.de/contact.

PontusHorn:
Using `meta.description` for excerpts is fair enough, but it might be a good idea to run it through Twig's `striptags` filter when printing it in `<meta name="description" />`, to allow for some basic formatting of the excerpt.

fourtonfish:
I'll feel pretty stupid if this has already been implemented, since I don't feel like I ever have the time to look through the changes -- and I also don't know if this is the best place to ask for feature request, but would it be of interest to add such a behavior that header tags (`h1

PhrozenByte:
@Lomanic, I casually found (and fixed) a bug in `PicoDeprecated` that could lead to your problems with 3rd party plugins, but I can't confirm that without a copy of your Pico installation. Please try it yourself.

@all, I'll update the docs in the course of this week, so I would like to call you up for testing and further feedback. @theshka, I then recommend to merge this PR in the next week, this unfortunately seems to be the only way to get the necessary attention to receive more feedback.

PontusHorn:
I've been using this for two small sites for about two weeks without any issues. No errors logged despite regular traffic. I'm using two plugins of my own making which have been updated to work with this version - haven't done any testing worth mentioning with plugins that haven't been updated.

index.php
```diff
@@ -1,15 +1,3 @@
 <?php
-
-define('ROOT_DIR', realpath(dirname(__FILE__)) . '/');
```
dav-m85: What is the purpose of moving those constants in a separat file ? Why not moving them to the Pico class constructor for example ?

PhrozenByte: The `ROOT_DIR`, `LIB_DIR` and `VENDOR_DIR` constants are required before calling the constructor, so it's impossible to move them there.

Splitting constant definition (`global.php`) and instantiation of Pico allows developers to integrate Pico into their own ecosystem and overwrite the constants as necessary. It allows e.g. the following:
```php
<?php

class MyApplication {
    protected $pico;

    // snip

    protected function initPico() {
        // either include the global.php
        require_once(PICO_DIR . 'global.php');
        // or define the constants on your own

        $this->pico = new Pico();
    }
    
    // snip
}
```

Anyway, this is no good solution at all, I focused too much on the existing code. I'll overhaul this, probably by replacing the constants with something better... Thanks for bringing this to my attention.

lib/Pico.php
```diff
@@ -0,0 +1,945 @@
...
+     *
+     * @return void
+     */
+    protected function readPages()
+    {
+        $pages = array();
+        $files = $this->getFiles($this->getConfig('content_dir'), $this->getConfig('content_ext'));
+        foreach ($files as $i => $file) {
+            // skip 404 page
+            if (basename($file) == '404' . $this->getConfig('content_ext')) {
```
PhrozenByte: Like in Pico 0.9, we exclude all `404.md` files in any folder (e.g. `sub/404.md`). Pico 0.9 however always uses the "root" `404.md`. What do you think about supporting directory-specific 404 files?

lib/Pico.php
```diff
@@ -0,0 +1,945 @@
...
+            unset($rawContent, $meta);
+
+            // trigger event
+            $this->triggerEvent('onSinglePageLoaded', array(&$page));
+
+            $pages[$id] = $page;
+        }
+
+        // sort pages by date
+        // Pico::getFiles() already sorts alphabetically
```
PhrozenByte: Like in Pico 0.9, files are strictly sorted in alphabetical order, so `sub/bar.md` and `sub/foo.md` are listed before `sub/index.md`. What do you think about making an exception for index files, so `sub/index.md` is sorted by `sub` instead?

lib/Pico.php
PontusHorn: Personally, I don't see a particularly compelling use case that makes up for the added complexity that would bring.

lib/Pico.php
PontusHorn: This seems like a good improvement. I'd say it makes more sense in most cases. For other cases, there's always the plugin hook onPagesLoaded.

lib/Pico.php
PhrozenByte: One could want to use a different `404.md` for his blog, e.g. showing a tag cloud, what wouldn't make much sense for non-blog-pages. Apart from the use case, should we then still exclude all `404.md` files or just the "root" `404.md`?

index.php
```diff
@@ -1,15 +1,3 @@
...
-define('VENDOR_DIR', ROOT_DIR . 'vendor/');
-define('PLUGINS_DIR', ROOT_DIR . 'plugins/');
-define('THEMES_DIR', ROOT_DIR . 'themes/');
-define('CONFIG_DIR', ROOT_DIR . 'config/');
-define('CACHE_DIR', LIB_DIR . 'cache/');
-
-define('CONTENT_EXT', '.md');
-
-require_once(VENDOR_DIR . 'autoload.php');
-require_once(LIB_DIR . 'pico.php');
```
theshka: Class files should exactly match the class name. `require_once(LIB_DIR . 'Pico.php');` :)

theshka:
> @theshka, I then recommend to merge this PR in the next week, this unfortunately seems to be the only way to get the necessary attention to receive more feedback.

@PhrozenByte yes, I think so as well, this is looking excellent! ... The change log and docs will need updating still, yeah?

PhrozenByte:
@theshka: Yes, changelog and docs still need a update. I'm currently working on some more improvements and will then update the docs accordingly. I think this PR can be merged in the course of the next week - as long as nobody gives feedback that needs action (@all, feedback is still appreciated!). I think this isn't in a hurry, a week sooner or later isn't important.

index.php
PhrozenByte: Please take a look at fc7632b and cdef7a6; any feedback?

bricebou:
@PhrozenByte sorry for the delay but I'd like to react to your comment https://github.com/picocms/Pico/pull/252#issuecomment-138115044 

Maybe I don't really understand what you say about registering new meta variables but I don't see how you don't need `isset()` checks. For example, with the Leaflet plugin or the Tag plugin, you still have to check if the variable is set for each page. But I may misunderstood... :/

PhrozenByte:
@Lomanic: Thank you for providing your Pico install! I've fixed all problems related to Pico - as far as I can tell everything else works now as expected. :smiley:

The remaining `Warning: array_unshift() expects parameter 1 to be array, null given in .../vendor/erusev/parsedown-extra/ParsedownExtra.php on line 28` error is related to your downgraded `parsedown-extra` version. As far as I can tell it seems that your downgraded `parsedown-extra` version is incompatible to the current version of `parsedown`. Newer versions of `parsedown-extra` check for `parsedown` >= 1.5.0, so you could try a 1.4.x release. Otherwise please refer to @erusev.

@bricebou: My wording was a little bit ambiguous. Starting with Pico 1.0 you don't need `isset()` checks anymore. Please see [lib/Pico.php#L630](https://github.com/picocms/Pico/pull/252/files#diff-e1ef3619f8e68d43f7065af5939e1dc8R630) and [lib/Pico.php#L642](https://github.com/picocms/Pico/pull/252/files#diff-e1ef3619f8e68d43f7065af5939e1dc8R642).

lib/Pico.php
PhrozenByte: I think the added 10 LoC is it worth to resolve this undocumented and not expected behavior...

bricebou: What is the unexpected behaviour you are referribg to?

Le dimanche 4 octobre 2015, Daniel Rudolf <notifications@github.com> a
écrit :

> In lib/Pico.php
> <https://github.com/picocms/Pico/pull/252#discussion_r41097723>:
>
> > +    {
> > +        return $this->content;
> > +    }
> > +
> > +    /**
> > +     * Reads the data of all pages known to Pico
> > +     *
> > +     * @return void
> > +     */
> > +    protected function readPages()
> > +    {
> > +        $pages = array();
> > +        $files = $this->getFiles($this->getConfig('content_dir'), $this->getConfig('content_ext'));
> > +        foreach ($files as $i => $file) {
> > +            // skip 404 page
> > +            if (basename($file) == '404' . $this->getConfig('content_ext')) {
>
> I think the added 10 LoC is it worth to resolve this undocumented and not
> expected behavior...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/picocms/Pico/pull/252/files#r41097723>.
>


-- 
Brice Boucard

PhrozenByte: See the commented line and 77f9390 :smile:

Pico 0.9 excluded all `404.md` files in any directory from the pages list, even they weren't functional. This never was documented and is unexpected behavior - you either expect that only the global `404.md` file is removed or that Pico supports per-directory `404.md` files. The first solution would break BC, so I decided that the additional 10 LoC are it worth to resolve this problem.

theshka:
@PhrozenByte could you chime in on this thread? https://github.com/bricebou/pico_ganalytics/pull/1
I'm not sure what I'm dong wrong with respect to the enabled/disabled state of plugins. @bricebou

Lomanic:
@PhrozenByte I confirm that you solved the issues with my legacy themes and plugins (last commit tested: 27d6946), I just had Notices/Warnings that I could easily disable :+1: . Now I think I will have to upgrade the plugins I use so they don't rely on the 00-PicoDeprecated.php shim, I will probably wait for your branch to be merged in master.

index.php
dav-m85: Well that is roughly what I had in mind when spoting these constants. Looks tidy now :)

index.php
```diff
@@ -1,15 +1,9 @@
...
-$pico = new Pico();
+require_once(__DIR__ . '/vendor/autoload.php');
+$pico = new Pico(
+    __DIR__,
+    'config/',
```
dav-m85: I wonder if we could move configuration directly in index.php, instead of maintaining a separate configuration file (or at least code the configuration retrieval here).

Let me elaborate. I'm using Pico within a Silex installation, redirecting all calls to a specific pattern to the Pico->run function. Thing is Silex has already its configuration system, and having Pico duplicates a bit the logic (I need to maintain/deploy one file for it).

Something cool would be:
```php
// index.php
$confArray = array(
    'someconf'=>array()
);
$engine = new Pico($confArray);
$engine->run();
```
... but of course it would oblige us to remove index.php from the repo, provide tempalte for ti and make things more complicated in the end.

One way to mitigate this would be to make the Pico constructor aware, if $config is array then I load it instead of fetching file.

Of course, this is a new feature and it would be totally ok to postpone it until 1.0 is merged. I can provide PR if the feature sounds usefull to you guys. @theshka @PhrozenByte ?

PhrozenByte: Providing an easy way to integrate Pico in other applications was the intention of removing the constants, so this definitely is a completely fair use case. It is not necessary to move the whole config handling to `index.php`, it's entirely sufficient to provide an API to do such things - so I've added a `Pico::setConfig()` method. See method docs for details.

PhrozenByte:
I've updated Picos inline docs, **feedback is appreciated** (click [here](https://github.com/PhrozenByte/Pico/blob/pico1.0/content-sample/index.md) to see the parsed version). As said, I'm no english native speaker, it would be great if someone reviews what I wrote. The inline docs should contain answers to any questions a typical **user** could have, so please check for completeness, too.

theshka:
I can go through and review spelling/grammar @PhrozenByte... Would you like them as line comments or pull-request on your fork?

PhrozenByte:
A PR would be great, thanks @theshka

dav-m85:
:+1: travis

PhrozenByte:
**Heads up!** I'll merge this PR on 6th November, the new release will be called `1.0.0-beta.1`. I would like to call you up for testing the new release, feedback is *always* appreciated!

Issue #258, opened 2015-09-29
Apache vconf


RuZniki:
Is it possible to add to documentation apache vconf example?

PR #264, opened 2015-10-08
Create apache config example file - Solve issue #258

dmelo:
- include an apache.conf.template inside config path, with instructions
about how to use it.

### DIFF
--- a/config/apache.conf.template
+++ b/config/apache.conf.template
@@ -1,0 +1,26 @@
+# Place this file inside /etc/httpd/conf.d and edit the fields. E.g.:
+#     cp ./config/apache.conf.template /etc/httpd/conf.d/pico.conf
+
+<VirtualHost *:80>
+    # fill in your email
+    ServerAdmin admin@yourdomain.com 
+
+    # replace by the real path
+    DocumentRoot /home/dmelo/Pico
+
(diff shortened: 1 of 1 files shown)


PhrozenByte:
In my opinion this is out of Picos scope, it's not our job to explain how a webserver is configured. Any other opinions about that? @theshka

theshka:
While I do agree it is out of Pico's scope @PhrozenByte, the Wiki does offer instructions for NGIX, and others. Perhaps we can include this in the wiki, or just remove them all and have examples for none?

PhrozenByte:
I think our configuration suggestions would be (per-definition...) wrong, there are too many platforms with different configuration strategies out there. For example, I'm using Debian, using the suggested template will definitely break Debians default configuration. It's even hard to link to a website that explains the webserver configuration, I don't know any website that even covers the major platforms, far less any platform. My personal opinion is, that someone, who isn't able to google for the right solution for his platform, shouldn't have a root server, he should stay with managed webspace. So, I think it's best to remove all configuration examples, except for Pico-specific things like URL Rewriting.

theshka:
Perfect, then lets move to clean up the examples, and going forward reiterate that web server configuration is not within the scope of this project. :+1:

Issue #270, opened 2015-11-06
Pico 1.1 and beyond


PhrozenByte:
The following points are just ideas and are supposed to stimulate a discussion about the future development of Pico. For the more distant changes/improvements planned for Pico 2.0, see #317.

* [ ] Allow Markdown configuration (`Parsedown` doesn't support much configuration, probably it will be limited to the optional usage of `Parsedown` instead of `ParsedownExtra`, `Parsedown::setBreaksEnabled()`, `Parsedown::setMarkupEscaped()` and `Parsedown::setUrlsLinked()`)
* [ ] Add only pages of the first level to the page navigation by default (e.g. `page.md` and `sub/index.md`, but not `sub/page.md`)<br/>*possible BC break, maybe we should defer this to Pico 2.0?*
* [ ] New default theme using Bootstrap 3 or (provided it has been released already) Bootstrap 4<br/>*possible BC break, maybe we should defer this to Pico 2.0?*
* [ ] Split `Pico::run()` into multiple methods, but still call them through `Pico::run()`
* [ ] Add `includePico()` function to prevent access to `$this`/`Pico` object in config/plugin files (use composer's `includeFile()` function instead?)
* [ ] Make `theme_url` configurable; the default value should support that `Pico::$themesDir` is a arbitrary deep sub path of `dirname($_SERVER['SCRIPT_FILENAME'])` (i.e. `index.php` resp. `httpdocs` folder); otherwise fallback to `themes` (and not `basename(Pico::$themesDir)`)
* [ ] Support multiple plugin dirs
* [ ] Add `onSinglePageLoading` event (new preliminary event to `onSinglePageLoaded`)
* [ ] When changing the `$page` parameter of the `onSinglePageLoaded` event to `null`, don't add this page to `$pages` (keeping `$page['sub/page'] = null` makes no sense, thus no BC break)
* [ ] Add `Pico::VERSION` constant

#### New official plugins
* [ ] Blog features
  * [ ] Pagination
  * [ ] Tags
  * [ ] RSS Feeds
* [ ] Editor/Page Admin
* [ ] Multilanguage (both contents and themes)

smcdougall:
I plan to take a crack at expanding the `cookbook` at some point (as well as the other docs), but I haven't had the time lately.  Unfortunately, my daily "productive time" went away mid December and won't be back until around the end of February.  I'd still like to *make* the time to work on it soon though.

[...]

PR #271, opened 2015-11-07
Use Twitter Bootstrap 4 for Pico default theme

theshka:
I know I am getting ahead of myself here, but this should take care of the suggestion in #270 to port the default theme to Twitter Bootstrap 4...

* [New] Require Twitter Bootstrap `4.0.0-alpha` via MaxCDN
* [New] added font-awesome stylesheet from MaxCDN …
* [New] add blog example to default theme …
* [New] blog, sidebar post, and full-width post layouts
* [New] blog/ directory in content-sample with 2 posts demo'ing layout
* [Changed] moved theme stylesheet to assets/css/
* [Changed] `style.css` => `bootstrap.theme.css`
* [Changed] Update `index.twig` to use bootstrap hirearchy/classes
* [Changed] demonstrate twig by splitting and including html snippets
* [Removed] `menu-icon.png`, bootstrap provides the text-based icon
* [Removed] `scripts/modernizr-2.6.1.min.js`, bootstrap provides `Normalize.css`


themes/default/index.twig
```diff
@@ -1,49 +1,70 @@
...
-    <link rel="stylesheet" href="http://fonts.googleapis.com/css?family=Open+Sans:400,700" type="text/css" />
-    <link rel="stylesheet" href="{{ theme_url }}/style.css" type="text/css" />
+    <!-- Page Title -->
+    <title>{% if meta.title %}{{ meta.title }} | {% endif %}{{ site_title }}</title>
+
+    <!-- Fonts -->
+    <link rel="stylesheet" href="http://fonts.googleapis.com/css?family=Open+Sans:400,700" type="text/css">
+
+    <!-- Stylesheets -->
+    <link rel="stylesheet" href="{{ base_url }}/vendor/twbs/bootstrap/dist/css/bootstrap.min.css" type="text/css" />
```
PhrozenByte: We forbid access to `vendor/` with our `.htaccess` by default. Probably a composer install script could solve this by copying the necessary assets to `themes/`. Any ideas?

themes/default/index.twig
```diff
@@ -1,49 +1,70 @@
...
-    </header>
+    </nav>
 
-    <section id="content">
-        <div class="inner">
-            {{ content }}
+    <!-- Main Content -->
+    <div class="container">
+        <div class="row">
+            <div class="col-xs-12 col-sm-12 col-md-12 col-lg-12 col-xl-12">
```
PhrozenByte: Isn't this obsolete when always specifying the full page width of 12 columns?

PhrozenByte:
Looks good :+1:

themes/default/index.twig
PhrozenByte: I've found a [good article about this problem](http://www.nikosams.net/blog/17_composer_npm_bower_assets_installation_using_composer-extra-assets) elaborating our possibilities. Additionally we can:
* write a simple helper script to copy the bootstrap assets to the desired directory
* add a whitelist entry to `.htaccess` (this will prevent users (like me) to put the vendor directory somewhere outside `httpdocs`)
* use one of Bootstraps CDNs

So we have at least 10 possibilites. Any thoughts @theshka, @dav-m85?

theshka: Thanks for the fresh eyes :eyes:  @PhrozenByte !

Since 4.0.0 isn't available on CDN yet, I opted to just create a `post-install-cmd` script in `composer.json` to copy the needed files to the theme directory, and therefore keep utilizing composers depenodency-managemnt.

themes/default/index.twig
theshka: Old habits die hard! You are definitely right, specifying just `col-xs-12` will apply to that one tier and all those above it. Changed.

composer.json
```diff
@@ -23,5 +24,11 @@
             "PicoPluginInterface": "lib/",
             "AbstractPicoPlugin": "lib/"
         }
+    },
+    "scripts": {
+        "post-install-cmd": [
```
PhrozenByte: https://getcomposer.org/doc/articles/scripts.md

> **Note:** Only scripts defined in the root package's `composer.json` are executed. If a dependency of the root package specifies its own scripts, Composer does not execute those additional scripts.

via #285

theshka: I suppose that when Pico 1.1.0 is ready to go, a Bootstrap 4 CDN will be available, if not, its just as simple to use Bootstrap 3. Then there will be no need to include with composer, or use the `post-install-cmd` script.

dav-m85:
Hahah this is serious cool stuff here. Let's get it merged !

theshka:
Closing this pull request. 

Moved the `feature/bootstrap-theme` branch from `theshka/Pico` to `picocms/Pico`

PR #273, opened 2015-11-13
Add map and sort_by Twig filters

PhrozenByte:
This is intended as a additional last-minute feature for Pico 1.0.0. Feedback is highly appreciated!

Especially the `sort_by` filter is important to allow users custom sorting algorithms apart from renaming their content files. This also makes Step 4 of our [Blogging section in the docs](http://picocms.org/docs.html#blogging) obsolete (it btw describes the problem a `sort_by` filter solves pretty well). Something like this is impossible to achieve with a Twig macro, so theme designers would otherwise very often need to also write a plugin. The `map` filter is just a by-product of the `sort_by` filter (better: of `PicoTwigExtension::getKeyOfVar()`).

What do you think?


theshka:
I like the proposed changes @PhrozenByte, it only adds functionality and thats great. :+1:  Perhaps including an example in the cookbook section of the docs and replacing the blogging instructions to use the new filter?

Issue #268, opened 2015-10-27
Pico 1.0 development & release process


PhrozenByte:
The phpDocumentor folder of the `gh-pages` branch (introduced with Pico 1.0 resp. #260) should be automatically generated when both a release is published (create a distinct phpDocumentor folder for any minor release - see http://semver.org/) and someone commits to the master branch.

By the way, our ToDo list for the final `1.0.0` release:
* [x] Add "Build & Release process" to `CONTRIBUTING.md`, added in bef1502
* [x] Automatically generate phpDoc, added in bef1502
* [x] Website: Add "What's new" to `upgrade.md`, added in #276 
* [x] Add a `README.md` for the `gh-pages` branch, added in https://github.com/picocms/Pico/commit/66bcb181cc82d760573a00d2bf472e4c7fb900c9
  * Don't forget to link to http://themes.iki-bir.com/webpaint/multipage/full/
* [x] Remove the `Pico 0.X Plugins` section from website with the final release of Pico 1.0.0
* [x] Update Pico Wiki
* [x] Wait for more feedback :smile:

Deferred:
* ~~New default theme using Bootstrap 3 or (provided it has been released already) Bootstrap 4~~
* ~~Website: Overhaul `_development/` collection resp. `development.md`~~
* ~~Website: Improve `_cookbook/` collection resp. `cookbook.html`~~
  * ~~Decide where to put links to it (inline user docs? website user docs? website dev docs?) and actually create the links~~

theshka:
What does our build and release pipeline look like? How should this be accomplished automatically?

PhrozenByte:
Actually I'm not sure about this, I think this can be achieved using one of those third-party tools (in GitHubs terminology "Webhooks, Services and [Integrations](https://github.com/integrations)"). In the past, either someone else did this or I did it manually so I don't have experience with that. I had the impression that maybe @dav-m85 has experience with it? Otherwise someone has to familiarize with this topic...

Manually the process looks more or less like this:
1. create & push a Git tag
2. run `composer` locally
3. create a ZIP archive from the resulting installation directory (so `vendor/` is included)
4. draft a release on GitHub (including the corresponding parts of the `changelog.txt`)
5. upload the ZIP archive as binary
6. run `phpDocumentor` locally, output goes to a new folder in the `gh-pages` branch
7. link the new phpDocs directory on the website
8. commit & push the changes to the `gh-pages` branch

theshka:
[...]

PR #278, opened 2015-11-18
Pico 1.0 development & release process

theshka:
Expanded the Build & Release section as per #268 so far. Still not sure where to announce the releases or if we even should. Is there anything missing? Incorrect?


theshka:
I was looking through some other repositories that auto-generate their documentation and upload to the gh-pages branch, and although I know for fact that this commit doesn't work, I feel like I'm on the cusp of something here... Could you take a look at https://github.com/theshka/Pico/commit/f062640bb1e87c82108067d238b47e7289ea11df @PhrozenByte?

PhrozenByte:
Sorry, I thought it would be faster when I make the necessary changes by myself instead of listing all steps. Great idea anyway :+1: It works just fine now :smiley:

theshka:
No need to apologize at all @PhrozenByte , thanks for cleaning up the idea. I'm just happy we can have all this documentation generation happen automatically for the next release :grin:

PhrozenByte:
What a ordeal... Complexity exploded... But works great. See [regular commit build](https://travis-ci.org/PhrozenByte/Pico/jobs/93274162) and [release build](https://travis-ci.org/PhrozenByte/Pico/jobs/93269616).

theshka:
:boom::exclamation: I think that it looks great @PhrozenByte ! 

You've made the script highly reusable, for every type of commit, branch and tags!, accounted for the potential race conditions, and tested that it works ... if you're confident with it -- merge at will. :smile:

PhrozenByte:
Thanks :smiley:
I actually didn't take a closer look to the doc changes yet, I'll catch up on this in the next days.

PhrozenByte:
I don't think it's necessary to describe the build & release process in such detail, especially not what Travis automates for us - one can look into the `.travis.yml` to find out the exact steps. A short paragraph like

> We're using [Travis CI](https://travis-ci.org/) to automate the build & release process. It generates and deploys [phpDoc](http://phpdoc.org/) class docs for new releases and on every commit to the `master` branch. Travis also prepares new releases by creating Pico's pre-built packages and uploading them to GitHub. You can trigger this process by pushing a new Git Tag, what references a (usually empty) Commit which message follows the following template: `...`

should be sufficient. The "Analysis phase" actually isn't relevant for the build & release process (why should one draft a new release on a previously failing build? :smile:). The manual steps should be in continuous text, the whole section seems so overwhelming otherwise. Btw, we always have to amend the release on GitHub, see https://github.com/picocms/Pico/issues/268#issuecomment-154492172.

PhrozenByte:
I've added [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer), it [basically tests code submissions](https://github.com/picocms/Pico/blob/enhancement/build-release/phpcs.xml) against PSR-2 coding standards. Any comments about it?

Thank you for your amendments, it's now very close to perfection! :+1: :smiley:

I've converted the Issues template to continuous text and slightly updated the build & release process section. I hope the changes are okay, can you please approve them? If this is the case, you can merge the PR afterwards. :smiley:

theshka:
I'm absolutely fine with the Code Sniffer checking for PSR compliance. :+1:  Out of curiosity, does the Travis build fail if it detects a violation, or does it correct it and move on?

PhrozenByte:
The Travis build fails (i.e. shows a warning, but doesn't correct anything), but still executes all subsequent commands of the `script` section. See https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Build-Step:

> When one of the build commands returns a non-zero exit code, the Travis CI build runs the subsequent commands as well, and accumulates the build result.
>
> In the example above, if `bundle exec rake` build returns an exit code of 1, the following command `bundle exec rake builddoc` is still run, but the build will result in a failure.

Because we don't have any subsequent commands in the `script` section, it actually stops after a failure, i.e. Travis neither updates the docs nor deploys new releases. This actually is intended, some checks of PHP_CodeSniffer are fatal (e.g. PHP parse errors), very serious (e.g. the deprecated or performance warnings) and I personally even consider PSR-2-violations as a contradiction to our "deployable" requirement when pushing to `master`.

PR #284, opened 2015-11-23
Update plugins page 

theshka:
Change `plugins` title to `add-ons`, and list all plugins and themes for Pico 0.X and Pico 1.0+ using the same togglebox as cookbook. Demo: http://theshka.github.io/Pico/plugins.html

### DIFF
--- a/plugins.html
+++ b/plugins.html
@@ -1,99 +1,121 @@
 ---
 layout: default
-title: Plugins
-headline: Pico Plugins
+title: Add-Ons
+headline: Pico Plugins &amp; Themes
 description: |
     Below are some plugins which extend the functionality of Pico and make it even more awesome.<br />
     Want to make a plugin? See [the docs](docs.html#plugins).
 nav: 4
 ---
+<h1>Plugins</h1>
+<div class="toggle">
+    <h4 class="title">... for Pico 1.0+</h4>
+    <div class="togglebox">
+        <div class="one-third">
+            <h3>Tags</h3>
+            <p>Provides tags functionality for Pico.</p>
+            <p><a href="https://github.com/PontusHorn/Pico-Tags" class="button red">Download</a></p>
+        </div>
+        <div class="one-third ">
+            <h3>Editor</h3>
+            <p>Markdown editor and file manager for Pico.</p>
+            <p><a href="https://github.com/theshka/Pico-Editor-Plugin" class="button red">Download</a></p>
+        </div>
+        <div class="one-third last">
+            <h3>Pico GAnalytics</h3>
+            <p>Google Analytics for Pico with basic options. </p>
+            <p><a href="https://github.com/bricebou/pico_ganalytics" class="button red">Download</a></p>
+        </div>
+        <div class="clear"></div><br />
 
-<h1>Pico 1.0 </h1>
-<div class="one-third">
-    <h3>Tags</h3>
-    <p>Provides tags functionality for Pico.</p>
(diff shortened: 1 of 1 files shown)


smcdougall:
Hmm.   It took me a minute to realize that I was supposed to *click* on the toggles.  Not sure how to make this more apparent.  I thought about putting a "+" in front of each header, but that looked awful (tried it using my browser's code inspector).

Another idea would be to [use tabs](http://themes.iki-bir.com/webpaint/multipage/full/tab-toggle.html) instead.  Have `1.0+` and `0.X` tabs under each section, with the `1.0+` being the default.  This would also prevent the page from looking blank when first loaded, while subtly guiding a user toward the `1.0+` content.

PhrozenByte:
The page shouldn't contain plugins/themes for Pico 0.9 at all

smcdougall:
Or that. :laughing:

smcdougall:
Also, you've made me realize how unflattering that `loosely based on the Anarcho Notepad theme` line is.  I'll probably remove that going forward, since the only remote way it's "based" on that theme is the similar styling and using the same (free) background. :unamused:

theshka:
I suppose in the not-so-distant future, the page will only display Pico 1.0 plugins & themes, but we haven't exactly dropped support for Pico 0.X yet, so why not have them featured at all @PhrozenByte? I liked the tabs idea if we are trying to be accommodating of prior plugins  :+1:

PhrozenByte:
With the release of v1.0.0-beta.1 I personally dropped support for Pico 0.9 :wink:

theshka:
Touché

PhrozenByte:
See 2a33837. We should remove the `Pico 0.X Plugins` section with the final release of Pico 1.0.0.

@smcdougall: You can give us a one-sentence-description of your theme and we'll add it to the website :smiley:

@theshka, regarding support of Pico 0.9: I don't see any advantages in supporting an old version of Pico which has known flaws and problems, or what do you think?

smcdougall:
For now, probably go with what I wrote on the wiki: "Dark wood grain, lined paper, and sticky note widgets."  If I think of something better I'll let you know or pull request it in the future.

As far as Pico 0.X support on the site, I think I agree with you (not that you asked for my input :stuck_out_tongue:).  A lot of the old plugins are likely unmaintained and a fresh start would be better than pointing to a bunch of incompatible plugins.  Besides, this is kind of a one time thing.  In the future there will likely be less breakage between versions.  Even if it feels strict right now to say that the previous versions are no longer supported, it's being done for the sake of making Pico better in the future.

Also, feel free to tell me to butt out. :wink: Just trying to give feedback where I can since it's usually just the two of you.

theshka:
Don't ever feel like your input isn't important @smcdougall! Pico hadn't been actively maintained or developed in almost a year. All of our contributions (no matter how small or large) are the only thing keeping Pico relevant-- so if you're interested in seeing Pico grow and flourish, comment away! ;)

PhrozenByte:
@smcdougall: Done
Apart from that I completely agree with @theshka's remarks. I know that my statements often seem to be a little bit "cold", but please don't be surprised about that, I'm no english native speaker, I try to do my best :wink:

smcdougall:
Your English is good enough.  If it wasn't for your profile saying Germany, I'd never have guessed. :wink: 

But don't worry, you've left enough friendly comments that when you seemed a little "cold" I figured you were just being blunt / to the point. :smiley:

Issue #285, opened 2015-11-23
Pico doesn't find vendor/autload.php when installed as composer dependency


dmelo:
I have Pico as a dependency on my project. It resides on `vendor/picocms/pico`. When `index.php` is called, I get an error saying that it cannot find `vendor/autoload.php`

PhrozenByte:
Hmm, I can't see a use case where this is actually a problem. When you use Pico as a dependency, why would you call `vendor/picocms/pico/index.php` directly? Wouldn't you rather copy it to your projects root directory (i.e. the `httpdocs` dir of your webserver)? Wouldn't you furthermore change Picos config, plugins and themes dir to bypass the need to change anything inside your `vendor` directory?

theshka:
When you put it that way, I would say that Pico's default `index.php` *is* intended for standalone installations @PhrozenByte. It's true, I probably don't fully understand what a project that includes Pico the way you described looks like... You say you don't see any use case where someone would call `vendor/picocms/pico/index.php` directly, I can't say one way or the other :confused: 

Why do you call it directly @dmelo?

dmelo:
I've installed pico and wanted to test if everything was working:

```
composer require picocms/pico
cd vendor/picocms/pico
php -S 0.0.0.0:8899
```

When I accessed http://localhost:8899 however, it was giving an error message. Saying it couldn't find `vendor/autoload.php`. Just thought it would be nice if Pico worked right away even installed that way.

For debugging, while I'm implementing my own `public/index.php` (which creates a instance of Pico), it is good to be able to compare it with a working instance (`php -S` on the `vendor/picocms/pico`).

PhrozenByte:
Hmm... Okay, basically this seems to be a valid use case.

Unfortunately I'm not very pleased with the idea of haphazardly looking for a `autoload.php` in `../../../vendor/` - a application should *never* assume anything about the directory structure below its own root directory, this can drastically go wrong and even cause security problems. Hence we need a way to safely distinct between a regular standalone installation and the use as library - but without heavily increasing complexity.

Just to be clear: We absolutely want to support the use of Pico as library, that's the reason why we specify `"type": "library"` instead of `"type": "project"` in our `composer.json`.

theshka:
While I agree it feels haphazard to assume anything about the directory structure above Pico, it appears that others deal with this issue similarly to my first thoughts in https://github.com/picocms/Pico/pull/286#issuecomment-159107251 @PhrozenByte 

See: 
[...]

PR #286, opened 2015-11-23
change Pico's default `index.php` to be compatible with `composer require picocms/pico`

dmelo:
I'm installing Pico as a dependency of a project. But it is hardwired to load autoload.php from project root. It doesn't work when it is inside vendor/picocms.

I'm not sure if there is a better way to address this issue.

It fixes issue #285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Released: -
           (RFC 3986) in `Page::evaluateRequestUrl()`
 * [Fixed] #272: Encode URLs using `rawurlencode()` in `Pico::getPageUrl()`
 * [Fixed] #274: Prevent double slashes in `base_url`
+* [Fixed] #285: Make `index.php` work when installed as a composer dependency
 ```
 
 ### Version 1.0.0-beta.1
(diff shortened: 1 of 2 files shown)


theshka:
I'm sure there's a more elegant way of doing it, but, we should probably check for composer locally and in the parent directory too before setting php configuration options, no?   

```php 
// load dependencies
if (file_exists(__DIR__ . '/vendor/autoload.php')) {
    // local composer install
    require_once(__DIR__ . 'vendor/autoload.php');
} else if (file_exists('../../../vendor/autoload.php')){
    // root composer install
    require_once('../../../vendor/autoload.php');
} else {
    // composer needs install...
    die('Cannot find composer `/vendor/autoload.php` -- try `composer install`');
}

//require_once(__DIR__ . 'vendor/autoload.php');

```

dmelo:
Is it better to check `file_exists` on both locations instead of fixing the set_include_path? I truly don't know.

There is no major improvement with this PR. It would just keep Pico working OOTB even if loaded as a dependency.

theshka:
I don't actually know @dmelo :confused: ... Discuss in #285?

Issue #292, opened 2015-11-29
`Unable to parse` when page meta has no "space after the two dots"


AeonFr:
I accidentaly broke my site for saving a wrong file. Simply putting `Author:Example` instead of `Author: Example` and the hole site becames unaccesible. This is what PHP throws:

```
Fatal error: Uncaught exception 'Symfony\Component\Yaml\Exception\ParseException' with message 'Unable to parse at line 2 (near "Author:Francisco Cano").' in ***/pico/vendor/symfony/yaml/Parser.php:298 Stack trace: #0 ***/pico/lib/Pico.php(764): Symfony\Component\Yaml\Parser->parse('Title: Ayuda\nAu...') #1 ***/pico/lib/Pico.php(958): Pico->parseFileMeta('/*\nTitle: Ayuda...', Array) #2 ***/pico/lib/Pico.php(337): Pico->readPages() #3 ***/pico/index.php(31): Pico->run() #4 {main} thrown in ***/pico/vendor/symfony/yaml/Parser.php on line 298
```

PhrozenByte:
Yes, this is a known problem. Unfortunately I've absolutely no idea how to work around this: What behavior would you expect? Silently ignoring it? The result would be that this page has no meta data, so e.g. the page title is empty (we could at least fallback to the filename), but I think hiding such problems makes the situation worse.

AeonFr:
I have the feeling this has something to do with the regex in parseFileMeta() function in lib/Pico.php
 
```
public function parseFileMeta($rawContent, array $headers)
    {
        $meta = array();
        $pattern = "/^(\/(\*)|---)[[:blank:]]*(?:\r)?\n"
            . "(.*?)(?:\r)?\n(?(2)\*\/|---)[[:blank:]]*(?:(?:\r)?\n|$)/s";
        # ...
   }
```
I can't understand a single character of that regex but it was in the stack trace.. :P

The problem broke my hole site, even the editor plugin. It would have been better if only that page would break (well, at least I could log in to picoeditor and fix it through the browser)  
I guess for now is just a matter of being careful about it

Thanks for the answer :raised_hands:

PhrozenByte:
No, it's not the regex, the exception is thrown by the Symfony YAML parser. Any thoughts about the expected behavior (also @theshka)?

theshka:
[...]

PR #294, opened 2015-11-29
Catch YAML parse errors (2)

PhrozenByte:
Possible solution for #292 (alternative solution: #293), catches all `\Symfony\Component\Yaml\Exception\ParseException` except for the current page and adds a `YAML_ParseError` meta variable with the error message. Therefore browsing to the erroneous page still shows the exception.

### DIFF
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -962,8 +962,7 @@ class Pico
                 try {
                     $meta = $this->parseFileMeta($rawContent, $headers);
                 } catch (\Symfony\Component\Yaml\Exception\ParseException $e) {
-                    $meta = array_fill_keys(array_keys($headers), '');
-                    $meta['time'] = $meta['date_formatted'] = '';
+                    $meta = $this->parseFileMeta('', $headers);
                     $meta['YAML_ParseError'] = $e->getMessage();
                 }
             } else {
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -971,9 +971,6 @@ class Pico
                 $meta = &$this->meta;
             }
 
-            // fallback to page id if page title is empty
-            $meta['title'] = (!empty($meta['title'])) ? $meta['title'] : $id;
-
             // build page data
             // title, description, author and date are assumed to be pretty basic data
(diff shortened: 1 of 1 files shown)


smcdougall:
Finally had a chance to test this and then realized... if the YAML error is stored as only a *meta* variable, it can't actually be used for anything because it would only be accessible on the page that's giving the exception.  I tried accessing it though the `page` variable, but I couldn't make that work either.

PhrozenByte:
Since Pico 1.0 the meta data of all pages is available through `$pages`, therefore you can read the error string via `$pages[...]['meta']['YAML_ParseError']`.

However, you can't do anything meaningful with it :wink: The only use case I can think of is highlighting the malicious page in a admin plugin...

smcdougall:
Yeah, I thought it should be available within `pages`, but I couldn't get it to work.  I was probably just doing something wrong.  Didn't really have the chance to properly focus on it, but I'll play around with it more soon.

I'm hoping to use it in a fashion similar to your example in #293, to help prevent `toc` errors (or other YAML errors) in [NotePaper](https://github.com/smcdougall/NotePaper).

smcdougall:
@PhrozenByte Yeah... I must have been really tired from the holidays because it's pretty clear in those comments that I wasn't even *thinking* clearly let alone coding properly.

Anyway, here's what I had in mind back then, finally coded.  It'll be in the next version of [NotePaper](https://github.com/smcdougall/NotePaper), which I should be done with soon.  I've been so busy cramming new features into NotePaper that I could have released several versions by now. :unamused: 

yaml error example

These warnings will appear on every page (except obviously the one that throws the exception when browsed to).  The idea is that it would be helpful to a user who has either edited a few files and doesn't know which one is causing trouble or who hasn't yet realized there's an error.

Thought I'd share, since to my knowledge no one else has implemented a use for your `YAML_ParseError` variable yet. :wink:

PhrozenByte:
Looks great! :+1: :smiley:

smcdougall:
Thanks! :smile:

PR #295, opened 2015-11-29
Isolate build-related files from source

theshka:
* [New] add `.phpdoc.xml` 
* [Changed] move phpdoc ignore list from script to xml 
* [Changed] rename `phpcs.xml` -> `.phpcs.xml`
* [Changed] rename `build/` dir -> `_build/`
* [Changed] update paths in `.gitignore`
* [Changed] update paths in `.travis.yml`
* [Changed] update paths in `.sh` scripts


_build/generate-phpdoc.sh
```diff
@@ -15,7 +15,7 @@ echo
 
 # generate phpdoc
```
theshka: With the addition of the code sniffer `.xml`, should we also have a `.phpdoc.xml` file in the root instead of hiding the ignore list inside the script?

PhrozenByte: Yes, definitely!

PhrozenByte:
@theshka, can you please look through the changes and merge the PR afterwards? Thanks! :smiley:

.gitignore
```diff
@@ -10,15 +10,17 @@ desktop.ini
...
-/build/phpdoc-*/
-/build/phpdoc-*.git/
-
 # Composer
 /composer.lock
 /composer.phar
 /vendor
 
+# phpDocumentor
+/_build/phpdoc/
```
dav-m85: If I get you well, you changed the build target directory to _build in order to avoir collision with the directory containing the phpdoc shell scripts.

Could we have rather moved the scripts to a "bin" directory and leave the build target dir untouched ? Hard to tell what is the difference between build and _build if you're not familiar with the project.

Just my two cents.

theshka: I proposed the prefix `_`change not to avoid collision with the shell scripts, but rather to signify that the directory is not apart of Pico's main files. Hope that clears it up @dav-m85 :smile:

PhrozenByte:
:+1:

PR #296, opened 2015-11-30
Refactor phpDoc auto deployment

PhrozenByte:
This should prevent empty phpDoc updates like 55b4a2b and furthermore simplifies the scripts


theshka:
:+1: the code is actually much more readable and verbose; I also noticed from last time that you added caching too!!! Awesome stuff here @PhrozenByte

PR #297, opened 2015-12-01
Use Twitter Bootstrap 4 for Pico default theme.

theshka:
Re-Pull of #271 - moved `feature/bootstrap-theme` from `theshka/Pico` to `picocms/Pico`
- [New] Use Twitter Bootstrap `4.0.0-alpha` 
- [New] Use Font Awesome `4.5.0` stylesheet 
- [New] `Blog Example` included with default theme …
- [New] `blog` (list), `blog-post` (post with sidebar) and `blog-post-full` (post full width) layouts 
- [New] `blog/` directory in `content-sample` with 2 demo posts using the new layouts
- [New] add `theme_config` to `config/config.php.template` for overriding default theme
- [Changed] demonstrate `twig` by splitting and including html snippets
- [Changed] use new `sort_by` twig filters 
- [Changed] moved stylesheets to `assets/css/`
- [Changed] renamed `style.css` => `bootstrap.theme.css`
- [Changed] update classes to use bootstrap hierarchy. 
- [Removed] `menu-icon.png`, bootstrap provides the text-based icon
- [Removed] `scripts/modernizr-2.6.1.min.js`


theshka:
I had originally planned on merging this with `v1.1.0`, however, I think this should be included in the final `v1.0.0`, and unless anyone has major objections or sees any glaring flaws, I would like to merge it soon.

PhrozenByte:
We're not in a hurry with this, the current default theme is not broken or something. Pico 1.1 will follow Pico 1.0 pretty soon. It's quite contradictory to tell people that Pico 1.0 is fully backwards compatible to Pico 0.8 and 0.9, and simultaneously changing any recognition value... A new theme btw breaks compatibility!

PhrozenByte:
What's the reason for closing this PR and deleting the branch? Postponement is not abandonment :wink:

theshka:
abandoned.
branch can be restored or copied from https://github.com/theshka/Pico/tree/feature/bootstrap-theme if you'd like to use anything from it for #270 :+1:

PhrozenByte:
Sure, I don't see any reason to start over, you did a great job, it's just a little bit too early to merge :smiley:

joshas:
Isn't Bootstrap a bit too heavy to be on default Pico theme? I'd like to suggest keeping current default theme as default, and maybe shipping Bootstrap theme as additional.

theshka:
That is how it currently is setup @joshas see: https://github.com/picocms/Pico/tree/feature/bootstrap-theme/themes

joshas:
Oh, this looks good :)

PhrozenByte:
@joshas, can you please elaborate on why you think Bootstrap is too "heavy"? Even it is currently implemented different, we're planning to use the Bootstrap theme as our default theme in Pico 1.1.

Bootstrap is widely known and a "fully featured" framework, something we cannot say about Pico's current default theme. Our default theme's concept is kind of "throw me away and build something own", Bootstrap gives users something to work with for their own website. With "Use Bootstrap" we don't wanna say that we will use a vanilla Bootstrap theme, @theshka's current work [already looks different](https://cloud.githubusercontent.com/assets/640217/11488304/1b17b388-978b-11e5-9827-532d14a79f2e.gif).

joshas:
I like the simplicity of current default theme. One html template, one styles file and absolutely no external dependencies (except fonts). By looking through index template new user can quickly figure out how themes work. Change a few lines of code, adjust some styles and you're good to go. Perfect starting theme for micro websites.

The proposed Bootstrap theme comes a bit of complexity. First: more files. One will need more time to get accustomed with it, find where's what. Second: Bootstrap adds about 100KB of its own styles and scripts. jQuery is required for some js components to work. Not everyone needs fully featured CSS framework to get a simple site running.

I use Bootstrap (as a whole and separate components) for a bit bigger projects and absolutely love the grid system. Work that @theshka has done is amazing. I believe that a lot of users will find Bootstrap theme useful and won't waste time searching for another CSS framework just to get started.
I only hope, that current theme will be included in default distribution, at least as a "blank" theme.

smcdougall:
This is a good point. I'd forgotten, but I did start my NotePaper theme by deconstructing Pico's default theme. It works *really* good as a example theme. Still, it's impracticable to use the default theme for anything but the most basic of websites.

I'm not sure what the solution would be though. Maybe have `content-sample` continue to use an updated version of the current theme while any user content uses the new theme. That wouldn't solve the "first-impression issues though"

Or a better idea, the current theme could be renamed "template" and included with the release but not used. The inline docs could then include information about switching to the template theme (or creating a copy of it, etc) to learn how theming works. This would be a similar concept to the plugin template that's already included in Pico.

PhrozenByte:
Hmm... Quite reasonable arguments.

First of all: **Templates**. I dare say you're right that we should keep the templates as simple as possible, i.e. spare includes and generally minimize the number of templates. Probably one or two templates are sufficient, e.g. a `index.twig` and a `blog.twig` to accomplish the desire for supporting blogging (at least basically). A complex template structure like the one we're using on our website definitely confuses beginners. This also increases customizability.

Second: **Stylesheet overhead**. Sure, Bootstrap has a large fingerprint. When assessing this, we should think about what we actually want to achieve first. Personally I'm completely focusing on the question: "How can we give users something to work with?". But what does that mean?

* Users need something with which they can layout their website. Bootstrap's grid system is a pretty good solution for that.
* Users shouldn't care about styling fundamental HTML elements, like headers, text (typography in general), lists, tables, code, images, etc. - basically everything Markdown supports without using HTML. Important to mention: Out of the box! This is something Bootstrap only provides partially, e.g. tables require a `.table` class to be styled. That's a problem...
* Lastly there are some features quite every website needs - like a navigation component, buttons and forms.

That's it. We don't need Bootstrap's JavaScript features like modals and dropdowns, we don't need badges, labels, progress bars or panels. Bootstrap actually allows [custom builds](https://getbootstrap.com/customize/), so reducing Bootstrap's fingerprint is possible - at least in theory. But this increases our problems about how we can keep Bootstrap up to date.

What do you think about that?

theshka:
The reason why I split everything in to `includes` is to have a collection of "building blocks" for users to include in custom layouts. I don't really see it being a problem having a couple variations of layouts and accompanying snippets as examples. Sure it is something new to learn, but Pico has introduced many new things in 1.X, including new default plugins, etc... 

As for keeping the default theme, why not? This could ship as the default "blog" theme, and have the current default theme ship as an auxiliary "bare" theme. 

@PhrozenByte mentioned, Bootstrap can be both simple and full-featured. Compiling a custom build and including only the minimal configuration would solve the remote dependencies concern as well...

PhrozenByte:
Starting with Pico 2.0 there will be a completely refactored default theme with heavily simplified templates and stylesheets (please refer to https://github.com/picocms/Pico/pull/334#issuecomment-237041996 and https://github.com/picocms/Pico/pull/334#issuecomment-237077916 for details). Pico's new default theme includes styles for all fundamental HTML elements, is very easy to extend and is a perfect starting point for users to create their own theme. I think this is sufficient for our goals (as elucidated above).

However, Bootstrap still is a widely known and "fully featured" framework many people would love to use. Thus I made the decision to create a official bootstrap theme with the typical Bootstrap + jQuery + Font Awesome troika before releasing Pico 2.0 (#334). Naturally we'll feature the theme on our website (http://picocms.org/themes/). Please stay tuned!

Thanks @theshka for your awesome work! 🎉 I'll use most of it for Pico's new official Bootstrap theme. Also thanks to all other commenters for their feedback! :+1:

PR #303, opened 2015-12-06
Changed 'content-sample' to 'content', revised some wording

smcdougall:

PR #304, opened 2015-12-06
Changed content to match docs changes.

smcdougall:
Merged the changes from #303 into this file.  Tried to preserve all the differences (such as escaping characters and the links in the table), but let me know if I've missed any.

Also, is there a reason this file has manual line breaks?  It displays in Pico about the same either way, so the only real difference would be if someone was viewing it in a text editor without word wrap.

I'd prefer to remove them to keep this content closer to what's in the Docs.  Right now it's inconsistent, with the new section not having any manual line breaks.  Also, removing the manual line breaks would make it look a little cleaner on Git / GitHub, which currently tells me I've made 30(!) Deletions instead of 10-13 or so.

Finally, it would be nice to either have a way to generate this file automatically or to mark the sections that must be changed between the Docs version and the Index version (extra tags added, escaped characters, links, etc).

As usual, let me know any changes you'd like to see to this file or #303 before merging them and I'll get it done. :smiley:

### DIFF
--- a/content-sample/index.md
+++ b/content-sample/index.md
@@ -10,18 +10,11 @@ Congratulations, you have successfully installed [Pico](http://picocms.org/).
 
 ## Creating Content
 
-Pico is a flat file CMS, this means there is no administration backend or
-database to deal with. You simply create `.md` files in the `content-sample`
-folder and that becomes a page. For example, this file is called `index.md`
-and is shown as the main landing page.
+Pico is a flat file CMS. This means there is no administration backend or database to deal with. You simply create `.md` files in the `content` folder and those files become your pages. For example, this file is called `index.md` and is shown as the main landing page.
 
-If you create a folder within the content folder (e.g. `content-sample/sub`)
-and put an `index.md` inside it, you can access that folder at the URL
-`http://example.com/pico/?sub`. If you want another page within the sub folder,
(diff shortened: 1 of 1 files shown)


smcdougall:
Actually, one idea could be to use twig comments to hide the changes inside the docs version that way if you copied over `index.md`'s content, you wouldn't be clobbering the changes you needed to preserve. You'd just have to remember to swap out or uncomment the changes.

`{# some changed text or tags #}`

PhrozenByte:
As our `CONTRIBUTING.md` states, the inline user docs *must* follow a hard line length of 80 characters. This improves readability in text editors. This inconsistency to the user docs on our website is due to technical reasons. A automatic syncing technique isn't desirable, the inline user docs should be an abstract of a much more extensive user doc on our website, what is just currently missing. Thanks for your PR anyway :+1:

smcdougall:
Okay, I'll get that updated shortly. It doesn't matter as much to me, it's just easier to edit without the line breaks. :wink:  Either way, I'd rather see it consistent. I just wanted to ask before I formatted it one way or the other.

What about the content? I tried to make a few sections more readable, but only in that first section I was working on.

At some point, I'd like to look over *all* the docs and clean them up accordingly. :smile:

PhrozenByte:
Thank you @smcdougall :+1:

Issue #308, opened 2015-12-23
Explain GitHub labels


PhrozenByte:
`CONTRIBUTING.md` should explain what the labels we're using on GitHub mean, like:
- Issues/PRs  without a `status` label are implicitly supposed to be "active" right now (e.g. under development)
- Issues/PRs without a `pri` label implicitly have "normal" priority
- `status: Pending merge` is used by PRs whose development is assumed to be finished
- `type: Discussion`, `type: Notice`, `type: Question` and `type: Idea` never have an `status` label
- `type: Discussion`, `type: Invalid`, `type: Notice`, `type: Question` and `type: Release` never have an `pri` label
- `type: Invalid` is for invalid Issues (like false bug reports); whether it was possible to help the user with his problem anyway, the Issue may be labeled with `status: Resolved`
- `info: Upstream` is for things related to `composer.json` dependencies
- `info: Website`, `info: Meta` and milestones are mutually exclusive
- `info: Meta` relates to Travis/our build environment, CodeSniffer, phpDoc, `README.md`/`CONTRIBUTING.md` etc. - i.e. not release specific things which aren't related to our website
- etc.

Probably we should explain the labels on the basis of our `type` labels, e.g.
- Issues which describe faulty, not intended behavior of Pico, are labeled with `type: Bug`. They must always include a `pri` and `status` label.
  - However, `pri` and `status` label can be omitted for the default cases (see above)
  - `pri: High` should only be used for showstopper bugs or security issues, `pri: Low` as the case may be.
  - [describe `status` labels] `status: Rejected` is never used in conjunction with `type: Bug`; use `type: Invalid` instead.

PhrozenByte:
I think we should reconsider the use of the `pri: Normal` and `status: In Progress` labels to reduce the number of used labels per issue (i.e. remove these labels). They are some kind of "default", thus it imo makes sense to only tag an issue with a `pri` and `status` label when its priority/status isn't the default. @theshka, what do you think?

PR #310, opened 2015-12-25
CONTRIBUTING.md: Create "Labelling of Issues & Pull-Requests" section

theshka:
Proposed solution to #308
- List all available labels
- Describe the labels purpose and usage
- Specify which labels not to use in conjunction with other labels

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,6 +77,63 @@ Pico's actual development happens in separate development branches. Development 
 
 As soon as development reaches a point where feedback is appreciated, a pull request is opened. After some time (very soon for bug fixes, and other improvements should have a reasonable feedback phase) the pull request is merged and the development branch will be deleted. Trivial bug fixes which will be part of the next `PATCH` version will be merged directly into `master`.
 
+Labeling of Issues & Pull Requests
+----------------------------------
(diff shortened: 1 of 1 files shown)


PhrozenByte:
Great work! :+1:

However, for reasons of clarity and comprehensibility, I suggest a approach that is a little bit different. Instead of listing all labels from top to down, I think it is meaningful to describe just the `type` labels this way. All other labels change their meaning in conjunction with a `type` label - partially significantly.

Examples:
* The usage rules of `pri: High` are very strict for `type: Bug` (security and showstopper bugs only), but this isn't the case for `type: Idea`/`type: Enhancement`/`type: Feature`, where they can be used nearly highhanded.
* Whereas we never label a `type: Bug` with `status: Rejected` (we either use `status: Conflict`, `status: Won't fix` or `type: Invalid` instead), `status: Rejected` is completely fine with `type: Idea`/`type: Enhancement`/`type: Feature`
* `status: Won't fix` is exclusively used with `type: Bug`, `status: Pending Merge` only with `type: Enhancement` and `type: Feature` pull requests
* ...

There are many more rules like this, so this would result in a quite big list of limitations, exceptions and dependences, what makes it very hard to understand. I gave a rough outline in the second paragraph of #308. However, using a table instead of a list seems to be a good idea.

PR #315, opened 2016-01-14
Add static shields.io badge generator

### LINKED ISSUE #314: Replace shields.io badges
Due to the success and extensive use of shields.io, the service unfortunately experiences problems with GitHubs rate limiting. As a result, Pico's `License` and `Version` badges in the `README.md` aren't accessible pretty often.

* Replace the `License` badge ([License](https://img.shields.io/github/license/picocms/Pico.svg)) with a static image hosted on our website/the `gh-pages` branch
* Replace the `Version` badge ([Version](https://img.shields.io/github/release/picocms/Pico.svg)) with a (more or less) static image hosted on our website/the `gh-pages` branch. Extend `_build/deploy-phpdoc-release.sh` to also update this image with `https://img.shields.io/badge/release-v$TRAVIS_TAG-blue.svg` (e.g. [Version](https://img.shields.io/badge/release-v1.0.0-blue.svg)) along with the phpDocs class docs
* Replace the `Build Status` badge ([Build Status](https://img.shields.io/travis/picocms/Pico.svg)) with [Build Status](https://api.travis-ci.org/picocms/Pico.svg)
* Replace the `Scrutinizer Code Quality` badge ([Scrutinizer Code Quality](https://img.shields.io/scrutinizer/g/theshka/pico.svg)) with [Scrutinizer Code Quality](https://scrutinizer-ci.com/g/theshka/Pico/badges/quality-score.png?b=master)

PhrozenByte:
Resolves #314


PhrozenByte:
Due to some [strange TLS bug in Ubuntu 12.04 LTS](https://bugs.launchpad.net/ubuntu/+source/openssl/+bug/861137) used by Travis CI, we cannot access shields.io using HTTPS. Downgrading to SSLv3 would have avoided this problem, but SSLv3 is disabled due to the [POODLE attack](https://en.wikipedia.org/wiki/POODLE). shields.io requires HTTPS. At the moment I don't have any idea how to solve this...

```
Downloading badge...
--2016-01-14 15:07:30--  https://img.shields.io/badge/release-test-blue.svg
Resolving img.shields.io (img.shields.io)... 104.18.44.160, 104.18.45.160
Connecting to img.shields.io (img.shields.io)|104.18.44.160|:443... connected.
OpenSSL: error:14094410:SSL routines:SSL3_READ_BYTES:sslv3 alert handshake failure
Unable to establish SSL connection.
```

PhrozenByte:
After playing around a little bit with a Ubuntu 12.04 LTS virtual machine, `curl` came in my mind... I've absolutetly no idea why `curl` works, but `wget` does not (it's an issue with OpenSSL and both use it the same way...), but anyway - it works!

See [Travis job #96.1](https://travis-ci.org/PhrozenByte/Pico/jobs/102712096#L454) and [`613a362` of PhrozenByte/Pico](https://github.com/PhrozenByte/Pico/commit/613a362cd2112413c644a81aaa08cc9223c15aab) for an example. As always, **feedback is appreciated!** :smiley:

Note to myself: Before merging this add both [the license](https://img.shields.io/badge/license-MIT%20License-blue.svg) and [initial release](https://img.shields.io/badge/release-v1.0.0-blue.svg) badge to `badges/pico-{license,version}.svg` of our website/`gh-pages` branch.

PhrozenByte:
I'm a little disappointed with the extent of feedback :disappointed:
We mean "feedback is appreciated" absolutely serious, this is no empty phrase! Pico is a community project!

smcdougall:
If it means anything, my feedback was going to be: "If it's a bug in Ubuntu and they don't feel like fixing it, I wouldn't bother working around it."

12.04 is an *old* release, and although it's still supported by Canonical, it's gotten to the point where they aren't paying much attention to it (especially with a new LTS due out in April).  Given that it's a bug in SSL, I'm surprised they wouldn't address it, but with the bug labeled "Undecided" importance, I wouldn't expect them to at this point.

On a side note, I just realized that bug was reported Sept 2011... over a month before development of 12.04 was even started... :confused:

Anyway, I didn't end up saying anything because I'm out of my league on these development issues and because you found an answer pretty quickly.  I would wonder about your fix though.  *Why* does it work?  Did someone put a workaround into curl causing it to fallback to SSLv3 automatically?  Might be worth investigating.  I'd like to say that anyone having issues should just upgrade to 14.04, but there's probably a lot of older web-hosts that haven't upgraded, etc.

I don't mind chiming up on issues like these... I just worry that I'll get annoying and/or not really be of use. :wink:

PhrozenByte:
Your feedback definitely means something! I don't think that feedback can ever get "annoying" or out of any use. Every feedback is helpful. It either endorses our current idea or forces us to view things different - what eventually encourages new, better ideas. Even if we didn't share the opinion ultimately, we took it into consideration. Feedback is a prerequisite for finding the best solution. Even when you can't add something concrete to the solution of a problem, you can still give feedback regarding the need for a fix or what your approach would be to find a solution. This is even true for :+1: or :-1: comments, they tell us whether we're on the right track or not.

In general, one can say that when there's a bildschirmfoto von 2016-01-26 14 40 04 label, you can't ever do anything wrong with commenting on it. This is actually also true for issues without this label, but with this label, we particularly ask for feedback. :smiley:

Maybe we should add these communication principles to our `CONTRIBUTING.md` :smile:

smcdougall:
Personally, I'm either anti-social and keep quiet, or I'm so talkative I'll reply every other comment.  There is no in-between. :laughing:

I think the feedback problem stems from the fact that there aren't many people watching the project right now.  Other than me, you, @theshka, and one or two others, I haven't really seen anyone else here.  There's been plenty of one-off issues being opened, and there's definitely a fair amount of Pico users... just not that many interested in it's development I guess.

The question becomes then... how do we raise awareness of the Pico *Project* and not just it's software.  I don't really have an answer for that.  As you'll remember, I stumbled in the door making an Issue or two.  I probably would have moved on just as quickly if I hadn't been asked to stay. :wink:

Issue #317, opened 2016-01-15
Pico's next generation plugin system


PhrozenByte:
This issue is intended to collect and discuss ideas on how we can make Pico's plugin system more flexible than ever. The changes break BC, therefore they [can't make it into Pico 1.1, but Pico 2.0](http://semver.org/).

### General
* Load plugins from `plugins/<plugin name>.php` (non-recursive) and `plugins/<plugin name>/<plugin name>.php` (the class file must be named after the directory) only. Plugins with multiple classes may need to register an autoloader on their own. Also see #203 
* Install Pico plugins through `composer.json`. I've no idea how to realize this at the moment...

### Event system
* Allow plugins to return `false` on preliminary events (e.g. `onContentLoading`) to prevent Pico from performing a specific processing step (`Pico::run()` skips `Pico::loadFileContent()`). Returning `true` or `null` works as with Pico 1.0 and changes nothing. The subsequent event is still triggered (`onContentLoaded`), but the payload variable (`$rawContent`) is empty. The event is triggered with special priority on this plugin (regardless of the regular processing order), so it can set the variable before any other plugin receives the event.
  * Example:
    * A markdown cache plugin returns `false` during `onMetaParsing` and `onContentParsing` to load both meta data and the parsed contents from its cache.
  * Affected Events:
    * `onContentLoading` (completely skips `on404Content…` events) and `onContentLoaded`
    * `on404ContentLoading` and `on404ContentLoaded`
    * `onMetaParsing` and `onMetaParsed`
    * `onContentParsing` (simulates `onContentPrepared`), `onContentPrepared` and `onContentParsed`
    * `onPagesLoading` (`onSinglePage…` events will be simulated) and `onPagesLoaded`
    * `onSinglePageLoading` (new event) and `onSinglePageLoaded`
    * `onPageRendering` and `onPageRendered`

* Allow plugins to return `false` on the `onRequestUrl` or `onRequestFile` events to completely skip Pico's processing. The only remaining event to trigger is `onOutput` (new event) right before Pico returns `$output`.
  * Example:
    * A static HTML cache plugin returns `false` during `onRequestFile`, bypasses Pico's processing completely and returns the cached contents during `onOutput`

### New official plugins
* Markdown cache
* Static HTML cache
  * Save rendered output of pages to static HTML files
  * Rely on OS to detect file changes (last modification time of `.md` files)
  * Add `Cache: No` meta header to prevent pages from being cached
  * Open questions
[...]

PR #318, opened 2016-01-23
parsedown extension

dav-m85:
I needed a way of extending the Parsedown extension, hence this PR. Not sure if it is the proper way though.

### DIFF
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -814,7 +814,9 @@ class Pico
      */
     protected function registerParsedown()
     {
-        $this->parsedown = new ParsedownExtra();
+        if (is_null($this->parsedown)) {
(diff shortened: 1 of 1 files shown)


PhrozenByte:
A plugin developer shouldn't be able to replace the Markdown parser with something different (what would be possible with a `Pico::setParsedown()` method). Unfortunately I don't have a ironclad solution either. You can find my current ideas in #317 (3rd item of "Event system"). I really appreciate your feedback about those ideas!

PhrozenByte:
We'll implement this feature by adding a new `onParsedownRegistered(Parsedown $parsedown)` event to Pico 1.1. Please refer to #334 for further discussions about this.

Issue #330, opened 2016-02-27
[Feature Idea] Modular Config


smcdougall:
Here's a brainstorm of ideas for improving Pico's config system and making it modular.

* Process all files in the config folder.
  *  Loading order shouldn't matter.  Any file should be of equal importance, and not nested together.
  *  Depending on how this system works, Pico's core config could either be loaded at any time or prioritized to load first.  If done right, I feel this shouldn't matter.
  * Rename Pico's core config to 'core-config.php' or whatever new file extension is decided on.
  * Make sure *not* to load any sample files.  Perhaps standardize the naming scheme of samples for future plugins and themes.  (For some reason, I really want to call 'plugins and themes' 'extensions').

  * Pros
    * This would allow users to update the config for their themes and plugins without touching their main Pico config.
    * It would also make *removal* of plugins and themes easier due to their config being stored separately and easily deleted.
  * Cons
    * This could result in unintentional loading of files (misplaced, misconfigured, or left behind).

* Switch Pico's config to another format.
  * Use something more user-friendly than PHP for configuration?
  * Preferably something more error-tolerant?  Something where a single missing comma cannot bring down an entire Pico site.
  * Despite format change, keep configuration just as powerful.  Make sure not to lose abilities, such as making arrays, that plugin and theme developers may rely on.

  * Pros
    * A new format could make Pico's configuration easier for new users.
    * It could make plugin and theme development easier.

  * Cons
    * Changing it could estrange developers, especially if they found it less powerful than PHP.

PhrozenByte:
This feature will be shipped with Pico 1.1, see 988a23f for implementation details. For feedback about this feature, please refer to #334. Thanks @smcdougall for bringing this up!

PhrozenByte:
**Update:** This feature will still be shipped with Pico ~~1.1~~ 2.0, however, with a different approach. See 9b7523b9e8fd418d9307208b8cde5b190703b1fb for implementation details. For feedback about this feature, please refer to #334. Thanks @smcdougall for further feedback about this there!

Issue #305, opened 2015-12-12
[Feature Idea] Pass $_GET to Twig template


smcdougall:
This seems like something that could be *immensely* powerful in templates.  I've already got two features for [NotePaper](https://github.com/smcdougall/NotePaper) that are planned around using GET variables in URLs.  This is however something that Pico doesn't do at the moment.

I was able to code this with one line in the `DummyPlugin`, by adding
`$twigVariables['TwigGetUrl'] = $_GET;`
to the `onPageRendering()` function.

While I could include a very basic plugin like this in my theme, it seems like kind of a waste, especially for something that could benefit others.

Best of all, since it's just passing a variable, it doesn't interfere with or change anything in Pico.  It does harmlessly pass along the page id used by the routing system, but that doesn't seem like it would really cause a problem (since most use cases would be searching for a specific item by name).

The only other issue I could think of would be that a theme developer would need to use the `rewrite_url` variable to determine whether to use a `?` or an `&` in their code (if they want to support both Rewriting and the Routing System anyway).

What do you guys think about this idea?  I'd love to see this as a feature, and it seems simple enough to do.

smcdougall:
https://github.com/picocms/Pico/commit/0c85d708204c72fc80dcfc07344036eaf10e4d08 Well that's certainly more than one line.  I take it you like the idea?  Yours is certainly more elegant than the `?` and `'?' ~ current_page.id ~ '&'` I'm using right now. :wink: 

Not sure how you plan to handle the *reading* of query data (most of your code is over my head), but it'd be neat to keep it in array format.  With the example above that I'm using, I have access to various bits of the query using `TwigGetUrl.queryitem`.  `?page=5` is accessable as `TwigGetUrl.page`.

Anyway, looking forward to seeing how this comes out. :smile:

PhrozenByte:
It wasn't intended that adding custom query data to an URL wasn't supported, so I've added a sane way to do this in 0c85d70. We probably will never support *reading* query data generically and without plugins, as this would introduce serious security issues (`$twigVariables[...] = $_GET;` unfortunately is no good idea). Input validation is hard - even for experienced programmers. It's nearly impossible with Twig. I'll reason this in detail when I have time for it, we can discuss the advantages and disadvantages after that. Sorry :unamused:

smcdougall:
:disappointed: 

[...]

Issue #299, opened 2015-12-03
[Feature Idea] Internal 404 document


smcdougall:
Now that my theme is complete, I've spent today rolling out `Pico 1.0.0-beta.2` onto my server.  I encountered two small issues during this, so I figured I'd open some Issues with my thoughts. :wink: 

I noticed that if there is no `404.md`, and your page 404's, Pico will throw an exception.  There's no mention of this in the Docs, it's purely suggested that `404.md` is an optional item.

I don't know about you, but I don't ever really modify `404.md`, because if I've done my job right, it will almost never show up.  I just use the default `404.md` provided in Pico's `content-sample`.

To me, `404.md` feels unnecessary.  I was thinking, why not bundle this page (or generate it) internally if it's not found in the content folder?  This way if someone wants to modify their 404 page, they can just like always, but if they don't include one, there's no ill effects.

PhrozenByte:
This feature will be shipped with Pico 1.1, see 00603f6 for implementation details. For feedback about this feature, please refer to #334. Thanks @smcdougall for bringing this up!

Issue #203, opened 2015-02-03
Should load_plugins() only load plugin_name.php?


coofercat:
I'm writing a plugin, and have found that any .php files in the plugin directory get loaded (by include()). I realise this is a change to the way Pico loads plugins, but I don't think it should do this. I'd favour only loading one file (probably a .php file with the same name as the plugin, eg. `plugins/my_plugin/my_plugin.php`). That one file should then do any include() it needs on any other PHP files.

My primary motivation is that I'd like to use an external PHP library to provide me with some functionality in some cases. Having Pico load it for me means it has to be loaded on every page view, which adds to response time and memory requirement unnecessarily. If my plugin can load it when required, then it only affects the page views that need it.

If this idea sounds useful/sensible/helpful, I'll produce a patch/pull request for it (all the fun is in the load_plugins() method, line 115 or so).

mistergraphx:
i've done this with a plugin.yaml description file im my case only the plugin with this file are loaded.

here is my modified function if it can help :

~~~~
/**
	 * Load plugins
	 *
	 * a plugin is minimaly described by a name in a plugin.yaml config file
	 * and a class/file . name, class and file must be identicals
	 *
	 * 
	 *
	 * @todo each web site must load only site.config activated plugins
	 */
	protected function load_plugins() {
            
		$this->plugins = array();
                
                $yamlFiles = $this->get_files(PLUGINS_DIR, 'plugin.yaml');
                
                if(!empty($yamlFiles)){
                    foreach($yamlFiles as $file){
                        $plugin =Yaml::parse(file_get_contents($file));
                        $plugin['path'] = str_replace('plugin.yaml','', $file);
                        $php_file = $plugin['path'].$plugin['name'].'.php';
                        
                        include_once($php_file);
                        if(class_exists($plugin['name'])){
                            $obj = new $plugin['name'];
                            $this->plugins[] = $obj;
			}
                    }
                }
	}


~~~~

Look how many pull-request are waiting , there is an issue about a new maintainer or active team for Pico. Nobody 'll never pull anything ...

PhrozenByte:
Ideas about how to achieve this in a backwards-compatible manner are appreciated! I though about this with Pico 1.0/#252 too, but didn't found a satisfying solution. Please also note that we don't want increase complexity ("stupidly simple").

PhrozenByte:
We'll very likely implement this with Pico 2.0, see #317 for further discussions.

PhrozenByte:
[...]

PR #334, opened 2016-03-03
Pico 2.0

PhrozenByte:
At the moment this PR is _"the mother of **WORK IN PROGRESS**"_... :smile:

Actually I don't have much to show yet, I'm primarily trying to point to [`PhrozenByte/pico-admin`](https://github.com/PhrozenByte/pico-admin). I decided to start with development of the admin plugin, because it may give us some more ideas for Pico ~~1.1~~ 2.0 *(renamed 2016-12-12)*. Please use this PR to give feedback about the plugin (otherwise the discussion may get a bit fragmented).

@theshka: Do you think it makes sense to ask @picocms once again to convert his/her account to an GitHub Organization? This will allow us to use [`picocms/pico-admin`](https://github.com/picocms/pico-admin) instead... This is particularly meaningful when bearing in mind that I'm planning to create multiple repositories (`picocms/pico-composer`, `picocms/pico-pagination`, `picocms/pico-tags`, `picocms/pico-feeds`, `picocms/pico-i18n`, `picocms/pico-search`)...

Most importantly we have to make a decision about replacing Pico's default theme (see #297) with Pico 2.0 or to defer this until Pico 3.0. What are your opinions about this? (especially, but not solely, @theshka) My objections are primarily related to BC, but I have not formed a opinion about this yet. Please also refer to the "Decisions to make" section below.

Refer to http://picocms.org/phpDoc/pico-1.1/ for updated phpDoc class docs.

Plannings for the more distant future (i.e. Pico 3.0) can be found in #317. ~~At the moment it's planned to let Pico 3.0 follow Pico 2.0 (i.e. no Pico 2.1).~~

As always, **feedback is highly appreciated!**

#edit: **This release has been renamed to Pico 2.0 (previously Pico 1.1) on 2016-12-12**, however we can't rename the branch, thus development will still happen on the `pico-1.1` branch.

#### Planned changes to Pico's core

- [x] Add `Pico::VERSION` constant
- [x] Add `includePico()` function to prevent access to `$this`/`Pico` object in config/plugin files (use composer's `includeFile()` function instead?)
- [x] #330: Implement a modular config
  - [x] Instead of using `*.config.php` files, use `*.yml` files to configure Pico. YAML is much easier to understand, more user friendly and (at least a bit) more error-tolerant, but still very powerful. Don't break BC by letting `PicoDeprecated` still read `config/config.php`.
- [x] Allow Markdown configuration (`Parsedown` doesn't support much configuration, probably it will be limited to the optional usage of `Parsedown` instead of `ParsedownExtra`, `Parsedown::setBreaksEnabled()`, `Parsedown::setMarkupEscaped()` and `Parsedown::setUrlsLinked()`)
- [x] Create a (de facto never updated) `pico-composer` project to allow `composer` users to install Pico as dependency and benefit from `composer`'s update mechanism; see @gilbitron's [BaunCMS/Baun](https://github.com/BaunCMS/Baun) repo
  - ~~Think about how users can easily install plugins with dependencies when they aren't using `composer`, but a pre-bundled release~~
- [x] #299: Integrated `404 Not Found` page
- [x] Add `Pico::is404Content()` method returning `true` when `Pico::$requestFile` doesn't exist (i.e. `on404ContenLoading` has been triggered)
- [x] Remove `PicoParsePagesContent`
- [x] Remove `PicoExcerpt`
- [x] Don't sort pages when a unknown `$config['pages_order_by']` is specified -- **breaks BC**
- [x] Trigger the `onMetaHeaders` event only once, cache results of `Pico::getMetaHeaders()`.
- [x] Update `README.md` to use a screenshot hosted on our website
- [x] Let users sort pages by arbitrary meta values (like Pico's `sort_by` Twig filter)
- Overhaul `Pico::parseFileMeta()`. Preserve BC with `PicoDeprecated`.
  - [x] Don't lower unregistered meta headers on the first level unsolicited (e.g. `SomeNotRegisteredKey: foobar` in the YAML Frontmatter should result in `$meta['SomeNotRegisteredKey']`, not `$meta['somenotregisteredkey']`).
  - [x] Don't compare registered meta headers case insensitive, rather allow registering multiple meta headers to result in the same meta value by flipping `Pico::$metaHeaders`
  - [x] Let users explicitly set the formatted date in the YAML Frontmatter (e.g. `date_formatted: Friday, 1st of December, 2000`)
- [x] Add a tree structure for all known pages

#### Planned changes to Pico's plugin system

- [x] Allow Pico plugins to trigger custom events
- [x] Add `onSinglePageLoading` event (new preliminary event to `onSinglePageLoaded`)
- [x] When changing the `$page` parameter of the `onSinglePageLoaded` event to `null`, don't add this page to `$pages` (keeping `$page['sub/page'] = null` makes no sense, thus no BC break)
- [x] Add plugin API versioning (new `DummyPlugin::API_VERSION` constant with Pico's `major`.`minor` version describing which API version the plugin expects; if the constant doesn't exist, assume `1.0`); this allows us to overhaul the plugin event system in the future without breaking BC
  - [x] #318: Add `onParsedownRegistered` event and pass `$parsedown` (instance of `Parsedown`) to allow extending `erusev/parsedown`
  - [x] Replace `onTwigRegistration` event with `onTwigRegistered` and pass `$twig` (instance of `Twig_Environment`), remove `$twig` parameter from `onPageRendering` event
- [x] Let `PicoDeprecate` differentiate between legacy levels (i.e. always enable the plugin to e.g. read `config.php` in Pico's root dir, but trigger deprecated events only when plugins of a specific legacy level are present)
- [x] Allow to manually load plugins (i.e. add `Pico::loadPlugin()` method)
- [x] Allow Pico plugins to be installed through `composer.json`
- ~~Support multiple plugin dirs~~
- ~~Add hook/event for `Pico::getPageUrl()`~~
- [x] Add more events to `Pico::readPages()`, similar to the events of the current page
- [x] Order plugins (execution order matters!) by evaluating their dependencies. Older plugins implicitly depend on `PicoDeprecated`. See [`tsort`](https://github.com/wertarbyte/coreutils/blob/master/src/tsort.c), [`marcj/topsort.php`](https://github.com/marcj/topsort.php), [`haberco/topological-sort-php`](https://github.com/haberco/topological-sort-php).
- [x] Add a event between `onPagesLoading` and `onPagesLoaded` passing the unsorted pages array (e.g. `onPagesDiscovered`).
  - Also remove the `$currentPage`, `$nextPage` and `$previousPage` parameters from the `onPagesLoaded` event, fire a separate event to "publish" them
- [x] Load plugins from `plugins/<plugin name>.php` (non-recursive) and `plugins/<plugin name>/<plugin name>.php` (the class file must be named after the directory) only. Plugins with multiple classes may need to register an autoloader on their own. Also see #203 -- **breaks BC**

#### Planned changes to Pico's UI (themes, Twig, routing, …)

- [x] #347: Drop the "index" part of URLs
- ~~Support multiple theme dirs~~
- [x] Make `theme_url` configurable; the default value should support that `Pico::$themesDir` is a arbitrary deep sub path of `dirname($_SERVER['SCRIPT_FILENAME'])` (i.e. `index.php` resp. `httpdocs` folder); otherwise fallback to ~~`themes` (and not~~ `basename(Pico::$themesDir)`~~)~~
- ~~Use Twig blocks in default theme (i.e. `{% block replace_or_extend_me %}...{% endblock %}`)~~
- [x] #305: Allow access to query string in Twig templates (i.e. `$_GET` superglobal)
- [x] Support `REQUEST_URI`-based URL rewriting to allow a much simpler webserver configuration
- [x] Use flexbox in Pico's default theme to pin the footer to the bottom of the window
- [x] Drop ~~`{{ site_title }}` and~~ `{{ rewrite_url }}` in favour of ~~their~~ its `{{ config.* }}` equivalent~~s~~. ~~Rename `{{ current_page }}` to `{{ page }}`.~~ Drop `{{ is_front_page }}` (use `{% if page.id == "index" %}` instead). Preserve BC with `PicoDeprecated`.  
- [x] Calculate the previous and next page for all pages (doesn't increase calculation time, `array_search()` in `Pico::discoverCurrentPage()` walks through all pages anyway). ~~Drop `{{ next_page }}` and `{{ previous_page }}` in favor of their resulting `{{ page.* }}` equivalents (see above). Preserve BC with `PicoDeprecated`.~~
- [x] Add only pages of the first level to the page navigation by default (e.g. `page.md` and `sub/index.md`, but not `sub/page.md`) -- **breaks BC** (more or less...)
- [x] Ignore all pages starting with a `_` for the page navigation by default (e.g. `_meta.md` and all files in `_collection/` are skipped) -- **breaks BC** (more or less...)

#### Website & Documentation

See picocms/picocms.github.io#18

- [x] Add Bountysource (merge `enhancement/bountysource` branch)

#### New official plugins/themes

The following plugins won't defer the release of Pico 2.0:

- Blog features
  - ☐ Pagination
  - ☐ Tags
  - ☐ RSS Feeds
  - ☑ #365: [File Prefixes](https://github.com/PhrozenByte/pico-file-prefixes)
- ☑ [Robots](https://github.com/PhrozenByte/pico-robots) (serve a `robots.txt` and `sitemap.xml`)
- ☐ Contact form
- ☐ https://github.com/picocms/Pico/issues/376#issuecomment-265972068: Excerpts
- ☐ #297: Bootstrap 4 theme
- ☐ [Editor/Page Admin](https://github.com/PhrozenByte/pico-admin)


smcdougall:
:+1: 

Also, glad to see some of my ideas in that list. :wink: :smile:

smcdougall:
I like the modular config implementation (https://github.com/picocms/Pico/commit/988a23fd024b065811185fca517939e59d561473) (the concept of it, can't really critique the php).

My only concern would be that `.config.php` is an odd naming convention / file extension.  There isn't really anything wrong with it, it just struck me as odd.

In practice, it would probably work well though.  `NotePaper.config.php` is just as good, if not better than `NotePaper-Config.php`.  :unamused:  It just seemed like a weird thing to *require*.

I suppose it'll help enforce a unified naming scheme between developers though. :laughing:

Anyway, nice work with it.  I was expecting a feature like that to take more than one commit! :+1:

PhrozenByte:
Updates (antedates/deferments from/to Pico 2.0, see #317):
* Antedate `pico-composer` project
* Antedate installing plugins via `composer`
* Antedate plugin API versioning (new `DummyPlugin::API_VERSION` constant)
* Antedate #318 by refactoring event system (also move `$twig` parameter of `onPageRendering` event to a dedicated event)
* Defer official Multilanguage plugin
* Defer official Search plugin

PhrozenByte:
New ToDos:
* Support independent (meta) data files (e.g. `content/catalog.yml`). The files are accessed the same way as pages (e.g. `{{ data.catalog.items }}`). If a data file has the same name as a markdown file (e.g. both `content/catalog.yml` and `content/catalog.md` exist), the data file is still accessible through `{{ data.catalog }}`, but its data is non-recursively merged into `{{ pages.catalog.meta }}` (YAML frontmatter takes preference). The same happens for all pages (non-recursive) in a directory if there's a data file with the same name as the directory (e.g. `_collection.yml` and `_collection/` directory). You can enforce recursion for e.g. `_collection/subdir/` by creating `_collection/subdir.yml`.
* [ ] Drop `{{ site_title }}` and `{{ rewrite_url }}` in favour of their `{{ config.* }}` equivalents. Rename `{{ current_page }}` to `{{ page }}`. Drop `{{ is_front_page }}` (use `{% if page.id == "index" %}` instead). Preserve BC with `PicoDeprecated`.
* [ ] Calculate the previous and next page for all pages (doesn't increase calculation time, `array_search()` in `Pico::discoverCurrentPage()` walks through all pages anyway). Drop `{{ next_page }}` and `{{ previous_page }}` in favor of their resulting `{{ page.* }}` equivalents (see above). Preserve BC with `PicoDeprecated`.

New decision to make:
* Implement the data files feature (e.g. `content/catalog.yml`) with a plugin instead
* Ignore all pages starting with a `_` for the page navigation by default (e.g. `_meta.md` and all files in `_collection/` are skipped)  
*possible BC break, maybe we should defer this to Pico 2.0?*

smcdougall:
> New ToDos:
> * Break EVERYTHING, but in the nicest possible way.

:wink:

Also, while I like the idea of having data files (extra possibilities are always fun), I wonder how useful they'd really be.  Right now, you'd have to utilize them in your theme to have them do anything useful, and I don't really see the advantage of that over using config variables (other than maybe easier formatting).

Now, if this were combined with something like `template_from_string` and some custom code on a specific page that acted on its extra data... well, it would behave a bit like Jekyll, wouldn't it... :unamused:

Still, it'll be interesting to see what could be built with these extra data structures. :smiley:

PhrozenByte:
One of the goals of this feature is to directly manipulate data from other services which don't necessarily understand the concept of a YAML frontmatter (e.g. the files in the [`_data` folder](https://github.com/picocms/Pico/tree/gh-pages/_data) of Pico's website are generated by [Shell](https://github.com/picocms/Pico/blob/master/_build/update-phpdoc-list.sh) [scripts](https://github.com/picocms/Pico/blob/master/_build/update-version-file.sh); Pico in contrast doesn't need a distinct `_data` folder). Config variables aren't supposed to be used to store "global" meta data - it's intended to use them to configure Pico's core and plugins (yeah, basically the configuration :smile:). Additionally, editing the config file (i.e. a PHP file) is a media disruption (something I'll possibly address in the future) and breaks the concept of "control the contents of your website by editing plain text files in the `content/` folder".

This feature neither breaks anything nor changes the way Pico behaves, it's a completely independent, additional feature which doesn't affect anything else. A user isn't forced to use it and it's not planned to break current approaches (wouldn't be possible without breaking Pico in whole anyway).

However, I actually already decided to implement it as a plugin. Probably not every user needs this and it can be implemented as a plugin, so the main prerequisites to add something to Pico's core aren't met. Additionally, I'll very likely drop the feature of treating data files named after directories specially.

All other new ToDos are basically to make a theme designer's life easier and drops legacy features (but still without breaking BC) :wink:

Btw: `{{ include(template_from_string(content)) }}` :stuck_out_tongue_winking_eye:

smcdougall:
When I said "Break Everything" I meant with the variable name changes.  Not sure if that was clear. :sweat_smile:

I was just trying to be funny. :laughing:

Anyway, it definitely all sounds interesting.  Honestly, I don't care what gets broken.  I'm the biggest fan of having "new stuff to play with".  I'd rather be on the bleeding edge of my software than be bored. :wink:

So... are you implying that `template_from_string` will be enabled? :smirk_cat:

PhrozenByte:
Ah, I see. Yeah, I'll change how `PicoDeprecated` works anyway, so this definitely will not break BC until you don't explicitly/manually disable `PicoDeprecated` :wink:

> So... are you implying that `template_from_string` will be enabled? :smirk_cat:

Oh, I overlooked that this needs to be enabled manually. No, I'm not planning to enable this by default.

smcdougall:
> Oh, I overlooked that this needs to be enabled manually.

Okay.  Yeah, a while back I wrote a modification for NotePaper replaced most `{{ content }}` tags with `template_from_string` versions and used a "plugin" (one or two lines in the DummyPlugin) to enable the feature.

It works surprisingly well, but the setup was over-complicated and I didn't actually need this feature, I was just making it for fun.  I felt that I was "messing with forces beyond my control" (or some other over-dramatic super-villain-y warning), and decided not to commit the feature.  After all, there must be a *reason* this isn't on by default, right? (Such as security concerns).

At some point I'll probably polish it up and release it as an *optional* feature.  It doesn't really benefit NotePaper much since NotePaper is geared toward written content and doesn't leave much wiggle-room for restructuring its layout.

Maybe Pico could offer `template_from_string` as an **advanced** config option?  Something we'd only recommend to power users who need (or expect) the ability to use twig in their content files.  People coming from Jekyll or other CMS's.

By default, Pico's content files should be for *content*, nothing else (thus, the default).  But content could mean different things to different people.  With some of the new YAML data options planned, `template_from_string` could provide some nice added utility.

Imagine that you had a large data-set you wanted to put in a table.  This data changes often, and you *really* don't want to have to edit an HTML table every time a change occurs.  Or maybe you want to have it sorted alphabetically, but not have to do this by hand every time you add data (mix in shifting table-rows, and that would be a nightmare).

By placing your table data in a matching `.yml` file, then generating it dynamically using twig, you could save yourself a lot of work. :smile:

I'm sure I don't have to go into hypothetical situations with you, *you're* the programmer after all. :wink:

What do you think of making this a toggle though? :confused:

PhrozenByte:
> Maybe Pico could offer `template_from_string` as an **advanced** config option? Something we'd only recommend to power users who need (or expect) the ability to use twig in their content files.

The problem of providing options for discouraged functions is, that users will use them - even when they don't know the implications. Many users actually need a "push" into the right direction. Therefore: No. However, it is still possible to enable this feature with a plugin. As a consequence, the discouraged function gets less attractive (because a user must take more steps) and the recommended alternative gains importance.

> Imagine that you had a large data-set you wanted to put in a table. This data changes often, and you really don't want to have to edit an HTML table every time a change occurs. [...] By placing your table data in a matching `.yml` file, then generating it dynamically using twig, you could save yourself a lot of work. :smile:

I'm not sure what the problem is, you're building the HTML table with a Twig for loop anyway!? The `Template` meta header exists for a reason. It's absolutely fine to create a template which is literally used by just a single markdown file. That's one of the reasons why themes should use includes to provide the basic layout - otherwise a user wouldn't be able to add custom templates. You will *never* cover all use cases of a all users with your theme - therefore you should make it easy for users to create custom templates.

> Or maybe you want to have it sorted alphabetically, but not have to do this by hand every time you add data (mix in shifting table-rows, and that would be a nightmare).

That's why Pico provides a `sort_by` Twig function... :wink:

smcdougall:
> It's absolutely fine to create a template which is literally used by just a single markdown file.

Yes, but that was my point.  You *could* do that, but it's a cumbersome and roundabout solution for a one-off layout.

> That's one of the reasons why themes should use includes to provide the basic layout

You've technically got a point there... :unamused:

> That's why Pico provides a sort_by Twig function... :wink:

You've missed the point here.  This is exactly what I was implying, but in the context of my argument, you would be using it in the `.md` file instead of a Template.

> Therefore: No.

One of these days I will successfully sway you in an argument. :triumph:

PhrozenByte:
> Yes, but that was my point. You *could* do that, but it's a cumbersome and roundabout solution for a one-off layout.

Don't get me wrong, your argument is completely valid! However, adding abstraction layers always leads to problems like that. By choosing a highly structured solution you always loose flexibility. Just think of Twig - we could use plain PHP instead (originally PHP wasn't more than a template engine). This is also true for software architectural patterns like [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) (Pico's content/theme separation is loosely based on the MVC pattern). Therefore, sure, you're completely right, however, the resulting disadvantages (we'll run into problems, just think of Jekyll and what a pain in the ass this content/theme mix is...) IMHO outweighs the advantages. And, as always, you can still achieve it with a plugin :smiley:

Btw, from my own personal experience: Initially you can't think of any other use case for it ("one-off layout"), but a year later you're using it for 10 other pages, too. :laughing:

> One of these days I will successfully sway you in an argument. :persevere:

You already did many times in the past, just think of #305 as one of the bigger examples or countless smaller things :wink: :smile:

smcdougall:
It's never a *clear* victory though. :stuck_out_tongue_winking_eye:

It's good that you've always got a clear focus of how you want things done though.

PhrozenByte:
Just because I'm reluctant doesn't mean that you didn't *clearly* win - it's the result that counts :stuck_out_tongue_winking_eye: Ideas without a concrete suggestion often need some time to have an effect, so don't underestimate your arguments :wink:

smcdougall:
That *is* why you supposedly keep me around, isn't it. :stuck_out_tongue_closed_eyes:

PhrozenByte:
[Pico Admin](https://github.com/PhrozenByte/pico-admin) just got a major update! However, Pico Admin is still just a developer preview and not yet usable...

bildschirmfoto am 2016-07-13 um 02 08 00

smcdougall:
> Pico Admin just got a major update!

Looking good!

Wow, you're building this from scratch? :+1:

At first glance, I thought it was built on the ever-forked Pico Editor. :laughing:
(A much improved version of it, of course :wink:)

PhrozenByte:
> Wow, you're building this from scratch? :+1:

Sure :wink: Pico Editor is very old, it doesn't benefit from Pico 1.0's plugin system and is a bit cluttered. Furthermore I feel like there should be a "bigger" plugin that demonstrates how plugins should be written. However, naturally I'm still depending on third-party projects like https://github.com/NextStepWebs/simplemde-markdown-editor

smcdougall:
Definitely!

And I think the sheer number of forks it has shows just how long that plugin has been on "life support" for.

I look forward to seeing the finished version.  Although I probably won't use it myself, I think it'll solve a barrier of entry for many users (especially those using a web host and/or shared hosting account).

---

Personally, I like my own setup.  I have all the "common files" of Pico symlinked to one Pico install on my server, with the unique files (content, config, assets) synced by Owncloud/Nextcloud.  It gives me local access to all my content, while automatically uploading any changes.

I'm just spoiled though. :sweat_smile:

PhrozenByte:
@all, I'm thinking of a simple way to implement a "Save As" dialog for PicoAdmin and would like to ask for ideas. On the one hand I don't want to create a full-featured dialog like in desktop applications, on the other hand adding just a text field sounds like a pretty inconvenient solution. The only idea I came up so far was to use the file navigation on the left - as long as the "Save As" dialog is open and a user clicks on one of the pages (e.g. `sub/page.md`), the path gets copied into the text field. One then just has to change the file name. However, I'm not really satisfied with this solution... Any other or additional ideas?

bildschirmfoto am 2016-07-21 um 00 21 08

smcdougall:
Just some random brainstorming, but what about having another copy of the file tree show as part of the save dialog.  Not a full featured file manager, just a copy of the tree view you already have.

You wouldn't be able to navigate anywhere, it would just provide a more intuitive place to click to grab the filenames from.

Also, you'd probably want to dim the rest of the page while the dialog is up in this case... otherwise the two identical file trees would look weird.

What do you think?  It would function the same as what you're already describing, but would hopefully feel more intuitive than having to click on the regular tree.

Most UI's train us that when a dialog box is open, we can't click on what's underneath.  I think that's the main problem with your suggestion.  It goes against that usual experience and that makes it feel "wrong".

PhrozenByte:
A full-page dialog is something I'm trying to avoid. I was thinking about integrating it completely into the navigation (i.e. no dialog at all), but didn't had a good idea about how to integrate it yet... I was thinking about letting a text field slide down as first item when hovering over an directory in the navigation (i.e. `sub.md` as representation of the `sub/` folder and `sub/sub` (there's a `sub/` missing :smile:)) or something like that...

smcdougall:
Well, that's up to you.  Just wanted to offer my two cents. :wink:

You said you were trying to avoid a "*full-featured* dialog", so my idea was a compromise.  It's *functionally* the same as what you've suggested, just presented differently.

I thought you were more worried about the logistics of a "full-featured dialog" (extra coding or dependencies, security concerns, integration with the other parts of the plugin, etc) than the appearance.

Personally, I think your version looks okay.  It reminds me a lot of Atom's file renaming box (and also its command palette).  But it does still need something.

It could be as simple as highlighting the navigation box to draw attention to it when the save dialog is up.  Something like a 1px, Pico-colored, border around the navigation tree.  That could very well look awful, but I thought it was a good example. :laughing:

PhrozenByte:
As you've said: We're just brainstorming. I'm thinking of both UX and amount of work.

Here's a (broken) mockup:
bildschirmfoto am 2016-07-21 um 16 18 50

As an alternative, I was thinking about dropping the "Create New" toolbar button (it's the first button) and forcing the user to specify the file name at the very beginning ("dropping" just means that it won't be in the default toolbar, one can still add it manually). I was thinking about adding a "Create New" button next to every page in the navigation (there's already going to be a small delete button when hovering over a page). When clicking on this button, the very simple notification-alike dialog (https://github.com/picocms/Pico/pull/334#issuecomment-234104177) shows up with the selected path pre-filled. The same dialog is going to be used when clicking the "Save As" toolbar button (pre-filled with the old path in this case).

I'm trying to make the whole process simpler while keeping the UI as simple as possible (minimalism is key). I just think that typing a full path is pretty inconvenient, thus I'm searching for some "reference point".

smcdougall:
As always, there's a line you have to balance between minimalism and usability.

I don't like the idea of dropping the "Create New" button, that just sounds like it'd be confusing. :confounded:

I think the goal should be to make the UX to be as *intuitive* as possible, even at the expense of some minimalism.  The idea of a "Create New" button next to every page is confusing.  If this were a "Duplicate page" button, that would be okay, but "Create New" is confusing.

Which "Create New" button should I click? Does it matter as long as it's in the same directory? Am I overwriting my existing file by clicking "Create New"??  These are questions I'd be asking as a new user (of the Plugin, not necessarily Pico).

If you're really set on using this "Notification" style dialog, I think your original idea is much better.  There's nothing really "wrong" with it.  If you hadn't asked for input, I wouldn't have thought anything bad about it.

> I'm thinking of both UX and amount of work.

As far as my proposal goes (the file tree dialog), I wouldn't think it would take much effort.  Of course, I don't know how your current code works either.

The idea was that you could reuse/call the same code that draws the file tree in the first place.  It would function exactly like you described in your first comment, https://github.com/picocms/Pico/pull/334#issuecomment-234104177, but would just be in a different place on the page.

smcdougall:
Also... I really *don't* like that green notification/dialog. :wink:

Grey was much sleeker looking.

smcdougall:
Also, I forgot about the "text-field" idea.  It's okay, but still not as good as the initial concept.

Instead of a text field, maybe a "Create New #" (replace `#` with a new page icon) button that appears at the top of the directory you're currently hovering over.  This would be significantly less confusing than the "next to every page" idea.

When a user clicks on this button, have the "Save As..." Dialog from https://github.com/picocms/Pico/pull/334#issuecomment-234104177 appear, populated with that location.  This would be labeled "Create New..." or something, not "Save As...".

Furthermore, with this setup, it could be used for "Save As..." as well.

When a user clicks "Save As", the dialog appears from https://github.com/picocms/Pico/pull/334#issuecomment-234104177, with the previously mentioned "click on the navigation" message.

But, the Create New buttons are also there now.  So if a user wants to save as a new file in a particular directory, they can click the corresponding "Create New" button to populate the field with the directory path, the same as before.

This example has kind of shifted to a "Create New" system instead of a "Save as" like you originally asked about... I liked where my idea was going though. :laughing:

The only thing I don't know about is how this "Create New" button would "push down" the directory contents without being annoying.  Maybe it should always be displayed, but faded instead of hidden.

Even with this "Create New" button, I think it's counterpart should still remain in the default toolbar.  Sometimes you just want to write without knowing if you're going to save it or not. :wink:

PhrozenByte:
> The idea of a "Create New" button next to every page is confusing. If this were a "Duplicate page" button, that would be okay, but "Create New" is confusing.
>
> I really don't like that green notification/dialog
>
> This would be labeled "Create New..." or something, not "Save As...".

You're going into too much detail for a *brainstorming*. How the button is being named or the color of the notification/dialog are completely unimportant details at this point... :wink:

> As far as my proposal goes (the file tree dialog), I wouldn't think it would take much effort.

It does. :wink: The file navigation is created by server, not by client.

> Instead of a text field, maybe a "Create New #" (replace # with a new page icon) button that appears at the top of the directory you're currently hovering over. This would be significantly less confusing than the "next to every page" idea.

Sounds legit, good idea. However, without a visible "Create New" label (i.e. just the icon) - there's simply not enough space for that.

> But, the Create New buttons are also there now. So if a user wants to save as a new file in a particular directory, they can click the corresponding "Create New" button to populate the field with the directory path, the same as before.

Hmm... Yeah, some sort of a always ("always" in terms of "as long as the dialog/notification is open") visible "Apply" buttons instead of "Create New" buttons next to directories...

> This example has kind of shifted to a "Create New" system instead of a "Save as" like you originally asked about...

"Create New" and "Save As" are intended to work the same way.

PhrozenByte:
As promised in 39d4a7c2aecd09fc8d4dcbe437ebd58b868e9da2: Here's Pico's new default theme (looks quite the same, however, with completely refactored source code). The mobile navigation is opened with a nice, smooth, CSS3-based animation (any modern browser and IE10+). The default theme generally works fine with IE9+.

---

**`README.md` Screenshot**

Pico 1.1 Default Theme

---

**Mobile View (Closed Navigation)**

bildschirmfoto am 2016-08-02 um 22 55 22

---

**Mobile View (Opened Navigation)**

bildschirmfoto am 2016-08-02 um 22 55 26

PhrozenByte:
**Screenshot of theme testing page**
bildschirmfoto am 2016-08-03 um 01 10 52

rbukovansky:
I'm :+1: x billion for this:

Add only pages of the first level to the page navigation by default (e.g. page.md and sub/index.md, but not sub/page.md)

Thanks.

PhrozenByte:
@rbukovansky, thanks for your feedback! Unfortunately I've to tell you that we'll pretty likely not ship this feature with Pico 1.1, but defer it until Pico 2.0. Pico follows [semantic versioning](http://semver.org/), what basically means that we never break BC (backward compatibility) with a 1.x release. Despite the fact that Pico's default theme actually isn't intended to be used for a productive website and more a starting point for your own theme, it still is a BC breaking change.

However, you can achieve this very easily yourself. Simply edit your Twig template and replace the navigation with something similar to the following:
```twig
{% for page in pages if page.title %}
    {% set pageDepth = page.id|split('/')|length %}
    {% if (pageDepth == 2) and (page.id ends with "/index") or (pageDepth == 1) %}
        {{ page.title }}
    {% endif %}
{% endfor %}
```

rbukovansky:
@PhrozenByte OK. Thanks!

PhrozenByte:
I'm thinking about renaming Pico 1.1 to Pico 2.0 (and the [currently planned Pico 2.0](https://github.com/picocms/Pico/issues/317) to Pico 3.0), primarily due to the fact that `PicoAdmin` requires the following BC-breaking change (strictly speaking; however, I don't think that it actually breaks something).

- Order plugins (execution order matters!) by evaluating their dependencies. Older plugins implicitly depend on `PicoDeprecated`. See [`tsort`](https://github.com/wertarbyte/coreutils/blob/master/src/tsort.c), [`marcj/topsort.php`](https://github.com/marcj/topsort.php), [`haberco/topological-sort-php`](https://github.com/haberco/topological-sort-php).

The problem I'm facing is, that the execution order of plugins matter: When a plugin hooks into the `onRequestUrl` event, it might do stuff a dependant plugin relies on. Thus it is crucial that events of the dependant plugin are always fired *after* all plugins it depends on. Up to this point it was mostly the user's responsibility to make sure that plugins are executed in the right order by adding numerical prefixes.

This change is imminent because the (not yet released) `PicoSession` plugin initializes the session during the `onRequestUrl` event, whereas the `PicoAdmin` plugin expects the session to be already initialized in this event. Thus `PicoSession` must be executed before `PicoAdmin` (what isn't the case due to the present alphabetical order). I don't want to add numerical prefixes because this problem isn't limited to `PicoAdmin` and `PicoSession`, it also applies to the admin modules concept which allows plugin developers to extend `PicoAdmin`.

Renaming Pico 1.1 to Pico 2.0 furthermore allows us to antedate some minor, but probably very useful BC-breaking changes (still I don't think that it actually breaks any existing plugin or theme):

- Add only pages of the first level to the page navigation by default (e.g. `page.md` and `sub/index.md`, but not `sub/page.md`)
- Ignore all pages starting with a `_` for the page navigation by default (e.g. `_meta.md` and all files in `_collection/` are skipped)
- Remove `PicoParsePagesContent` (add it to the plugin portfolio on Pico's website)
- Remove `PicoExcerpt` (add it to the plugin portfolio on Pico's website)
- Don't sort pages when a unknown `$config['pages_order_by']` is specified
- Load plugins from `plugins/<plugin name>.php` (non-recursive) and `plugins/<plugin name>/<plugin name>.php` (the class file must be named after the directory) only. Plugins with multiple classes may need to register an autoloader on their own. Also see #203
  - *I'm not sure about this one yet, as this might actually break some old plugins...*
- Trigger the `onMetaHeaders` event only once, cache results of `Pico::getMetaHeaders()`.

Besides: Pico's new default theme actually is also a BC-breaking change...

The additional programming effort isn't a problem, all changes are actually pretty minor.

Feedback is highly appreciated! (@all, specifically @smcdougall)

smcdougall:
> specifically @smcdougall

😆

---

Overall, I've got no opposition to renaming this release.  It's had very lofty goals from the beginning and always felt like a bit much for a *single* minor version bump.

The way you've described the plugin dependency issue seems like something that was once minor has snowballed into a bigger issue.  I don't see what the point of trying to work around it would be, since any fix would both be only temporary and somewhat hacked together.  It'd be a lot of extra effort for the sake of keeping the version number the same.

So, your logic is pretty sound.  It seems like it'd be a good idea to label this `2.0` going forward.  I mean... it's just a number after all. 😉 

And the other issues you mention seem like they'd be decent quality-of-life / UX improvements, so it's a good excuse and time to take care of those as well.

PhrozenByte:
Done :smiley:

PhrozenByte:
> - Load plugins from `plugins/<plugin name>.php` (non-recursive) and `plugins/<plugin name>/<plugin name>.php` (the class file must be named after the directory) only. Plugins with multiple classes may need to register an autoloader on their own. Also see #203

I haven't found a single plugin that depends on including *all* PHP files recursively. Most plugins consist of a single file anyway, just some plugins need multiple files - and just a few of them actually violate the desired `plugins/<plugin name>/<plugin name>.php` constraint. This issue can be fixed by simply renaming the directory to match the class name. However, I'll give the whole procedure some room for minor mistakes by matching directory and file name case insensitive (PHP treats class names case insensitive, too).

Definitely a BC break, but I think the problems are negligible. Any other thoughts about that?

smcdougall:
Nope, way over my head. 😉

PhrozenByte:
I'm thinking about changing the current solution for #330... Instead of reading `config/*.config.php` files, we might use `.yml` files instead; this includes removing even the main `config/config.php` in favor of `config/config.yml`. Needless to say that we don't break BC, `config/config.php` will still be read by `PicoDeprecated` (the same way we still read a `config.php` in Pico's root dir). As we all know, YAML is very powerful and allows us to do quite the same things as with PHP (except for things we never supported intentionally, like adding arbitrary PHP code). However, YAML is much easier to understand, more user friendly (Pico users must know YAML anyway) and (at least a bit) more error-tolerant (however, as we know from #292, it's definitely not perfect).

Any thoughts?

---

Furthermore I was thinking about the following two ToDos:

> - Support multiple plugin dirs
> - Support multiple theme dirs

Their only reason to exist is due to the fact that they solve two related ToDos:

> - Create a (de facto never updated) `pico-composer` project to allow `composer` users to install Pico as dependency and benefit from `composer`'s update mechanism; see gilbitron's [BaunCMS/Baun](https://github.com/BaunCMS/Baun) repo

When creating a `pico-composer` project, a user wants to install his/her plugins/themes into the "root" `{plugins,themes}/` dirs. However, by supporting just a single plugin/theme dir, we wouldn't be able to load Pico's default theme and plugins (i.e. `PicoDeprecated`) - they wouldn't be present in the "root" `{plugins,themes}/` dirs, but `vendor/picocms/pico/{plugins,themes}`. Supporting multiple plugin/theme dirs allows us to load plugins/themes from both locations.

> - Allow Pico plugins to be installed through `composer.json`. I've no idea how to realize this at the moment... Possible solution: [`composer/installers`](https://github.com/composer/installers)

When "installing" plugins and themes manually, one moves them to the `{plugins,themes}/` dirs. However, complex plugins with external dependencies (like Pico's official admin plugin) and for composer installations, one wants to use composer instead. By supporting multiple plugin/theme dirs, we can let composer install plugins/themes just regularly to `vendor/[developer name]/[project name]` and let Pico treat these dirs as additional plugin/theme dirs. However, there are still open questions (e.g. we deny access to the `vendor/` dir in our `.htaccess` by default, however, themes always include at least a CSS file that needs to be accessible).

Anyway, supporting multiple plugin/theme dirs might not be the best solution to these ToDos.

I was thinking about using [`composer/installers`](https://github.com/composer/installers), what allows us to install plugins/themes directly to Pico's `{plugins,themes}/` dirs instead of `vendor/[developer name]/[project name]`. `composer/installers` is a pretty common solution to do things like this, so this shouldn't be a problem at all. Thus we don't need multiple plugin/theme dirs to support composer-based plugin/theme installations.

However, this doesn't solve the problems we have with a `pico-composer` project. Due to the fact that `PicoDeprecated` is the only pre-shipped plugin (`DummyPlugin` actually is no "real" plugin) and Pico's default theme is the only pre-shipped theme, we might simply remove them from the `picocms/Pico` repo and use separate `picocms/pico-deprecated` and `picocms/pico-theme` repos instead. This allows us to specify them as dependencies for `picocms/Pico`, what effectively changes nothing compared to the current situation.

The reason why I don't want to support multiple plugin/theme dirs is, that it adds pretty much complexity to Pico. For example, what is the expected return value of `Pico::getPluginsDir()`? How does one influence the execution order of plugins? It simply isn't necessary as we have (as I think) better alternatives.

Any thoughts?

smcdougall:
I am definitely, 100% for using YAML-based config files.  When put together with reading all configs in the folder, I think it would be a huge UX improvement over having to paste custom lines into `config.php`.

I'd envision that a theme could ship with it's own custom `.yml` to be placed into `/config`.  Perhaps we should also call this the "recommended" method as well, because at the moment there's no standard way of adding variables.  Some themes provide a standalone `.php` config (NotePaper), others have lines to add to `config.php`.  Some themes use the YAML Frontmatter of `index.md` for variables (for example, other themes that @BesrourMS and I have ported).  It's kind of a mess... 😅

---

As far as the Composer stuff goes, I don't know much about it.  😒

As always though, simplicity is better.  Multiple theme/plugin folders sounds like a headache.

Also, separating `pico-deprecated` and `pico-theme` into separate repo's sounds like it could be handy.  Personally, I think these could be optional items.  Do we really have that much community content that still relies on the `pico-deprecated` plugin?

The release of `1.0` seemed to do a good job of clearing out old content.  90% of the items in the Wiki got updated, and most of the remainder aren't really relevant anymore.  Not sure if this will hold true for `2.0`'s release or not, but my point is just that I think those still interested (and up-to-date) with Pico are using modern add-ons.

Not that it weighs in too heavy at `15.9 KB`, but I think the idea of it being in a separate repo furthers the idea that it is only meant to be a workaround for older content, not a permanent solution.

Just my 2¢ though.... and at this point I'm rambling... 😉

PhrozenByte:
As discussed above and in preparation for [Pico's composer starter project](https://github.com/picocms/pico-composer), Pico's default theme has now a new home at https://github.com/picocms/pico-theme. This is also true for the `PicoDeprecated` plugin, which can now be found at https://github.com/picocms/pico-deprecated.

Please note that installing Pico will silently fail (at least until you navigate to the website) until composer/installers#359 is merged and released. This isn't crucial until we actually release Pico 2.0, so there's no big hurry. You can use symlinks (`ln -rs vendor/picocms/pico-theme themes/default` and `ln -rs vendor/picocms/pico-deprecated plugins/PicoDeprecated`) as a workaround.

PhrozenByte:
**Pico v2.0.0-beta.1 is finally here!** 🎉 ❤️ 

See #401

PhrozenByte:
**Pico v2.0.0-beta.2 is out now!** 🎉

See https://github.com/picocms/Pico/issues/401#issuecomment-359288809

PhrozenByte:
**Pico v2.0.0 is out now!** (yes, the stable release!) :tada:

See https://github.com/picocms/Pico/issues/401#issuecomment-401617414

PR #341, opened 2016-04-14
Nitrous.io Quickstart button

yeouchien:
Added Nitrous Quickstart files so anyone can quickly spin up a free, cloud dev environment to start using this repo. Clicking the Quickstart button will automatically clone this repo and will install and configure any required packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,17 @@ $ curl -sS https://getcomposer.org/installer | php
 $ php composer.phar install
 ```
 
+Nitrous Quickstart
+------------------
+
+You can quickly create a free development environment for this Pico project in the cloud on www.nitrous.io:
+
+<a href="https://www.nitrous.io/quickstart">
+  <img src="https://nitrous-image-icons.s3.amazonaws.com/quickstart.png" alt="Nitrous Quickstart" width=142 height=34>
+</a>
+
+Simply access your site via the `Preview > 3000` link in the IDE.
+
 Upgrade
 -------
 
--- a/README.nitrous.md
+++ b/README.nitrous.md
@@ -1,0 +1,9 @@
+# Setup
+
+Welcome to your Pico project on Nitrous.
+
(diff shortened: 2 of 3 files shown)


PhrozenByte:
Thank you for your PR @yeouchien! Can you please give us a brief overview about what nitrous.io is and what it can do for users and plugin+theme developers of Pico? What's the advantage for Pico's core developers and the Pico project in whole? Why should we promote such a service in general and why should we promote this particular project?

@picocms @theshka

yeouchien:
Hey @PhrozenByte ! 
We just want to make it incredibly quick and easy to contribute themes & plugins for Pico. The QuickStart button provides a way to create a sandbox Pico dev. environment with everything setup and installed to start building things right away.

We’ve found that reducing the friction in getting started helps considerably with engaging more developers to participate in open source projects. There’s also some really powerful “google-docs” style collaboration features so developers can pair program and help each other out. It’s 100% free for up to 2 projects.

All the best.

smcdougall:
I don't know, I don't have a good feeling about this.  It's one thing to support Pico on your platform. In itself, that seems like an okay idea.  It's another thing entirely to include a section about it in Pico's readme (or even on the website for that matter).  At that point it would appear not just to be supported, but *endorsed* by Pico.  Neither of which are really the case.

Furthermore, let's be honest here, this is advertising.  Plain and simple.  I don't think that Pico should be advertising for a service that it's not affiliated with, even if it did provide a good service.

I don't think there's any benefit to this either.  I feel like current users will be unlikely to need such a service, and may not appreciate being advertised to.  More importantly though, I feel like this could be confusing to new users.  With such a prominent location in the readme, new users may feel that this is the *recommended* option for setting up Pico, rather than just a third party service.

My vote is No.  If you want to include a Pico Quick Start option on nitrous.io, great, more power to your users, but I don't think the advertisement for it belongs *in* Pico itself.

PhrozenByte:
Thank you @smcdougall for elaborating this, I'm inclined to agree with you.
Unfortunately @yeouchien we can't accept your PR, thanks for your contribution anyway!

PR #342, opened 2016-04-16
Update style.css

ErickMR19:
Responsive design is currently broken in default theme when the width is between 768px and 850px, because the width of inner class is set fixed at 850px


PhrozenByte:
Thank you @ErickMR19!

Issue #343, opened 2016-04-16
Improve Nginx Documentation


smcdougall:
Pico's Nginx documentation is practically non-existent.  It should be improved and provide some detailed examples equivalent to what is provided for Apache in `.htaccess`.

A discussion has already begun on https://github.com/picocms/Pico/commit/d8f916691809b19b05104906ee3700c5cf3f8407.  It seems there is a lot to consider when using a Regex Location Block for Pico's rewriting, as Nginx will only use the first matching rule it finds.  Since a typical Nginx setup also uses Regex to detect when to send a page to PHP, a misconfiguration can easily break PHP detection altogether.

The solution seems to be placing the PHP Location Block *above* any Pico rewrite blocks, that way any requests to `index.php` get properly matched.

In the current example, `location ~ ^/pico(.*)` matches both the Page ID you're trying to rewrite (eg `/pico/page`) and also the index itself (at `/pico/index.php`, resolved from the URI `/pico/?page`).

Since `/pico/index.php` matches `location ~ ^/pico(.*)`, Nginx has succeeded in finding a match and will not move on to match it to `location ~ \.php$`.  The PHP page gets served as a download file instead of interpreted by PHP.

Robby-:
Hi, I'm the same Robby from IRC. While your proposed solution works, the solution I'm using is actually to include the PHP location block in each location block where I need it. To avoid having to duplicate this in too many places I put the PHP location block in a snippet file and include it in each location block as needed.

The reason I'm currently doing it like this is because of the following.

Not sure if this is useful for the situation described in this issue, but here is a problem I had:
The other day I was looking to password protect some specific directories of another PHP application with nginx and http auth and I found that PHP was still getting executed instead of requesting http auth.

The following links has relevant reading material on nginx and (nested) location blocks:
https://serverfault.com/questions/375961/nginx-basic-http-authentication-on-php-script
https://serverfault.com/questions/119076/how-to-use-fastcgi-globally-and-basic-auth-in-sublocations-in-nginx

Back to pico.
Note: I'm not using a subdirectory for pico anymore, I've put it in the root directory instead.
So, I noticed pico has an `.htaccess` file and I wanted to do some nginx equivalent of protecting for example pico's config directory, by doing this:
```
location /config/ {
  return 403;
}
```
But, the above does not block access to `/config/config.php` (a blank page is shown instead of a 403).

So, taking into account the information in the 2 links I mentioned above, the following location block does block access to `/config/config.php`:
```
[...]

PR #345, opened 2016-04-21
Nginx Documentation

smcdougall:
Rough draft is complete.  Still needs further revision and editing, but I wanted to share what I've gotten done so far.

View-able at http://smcdougall.github.io/Pico/nginx/

Any input is appreciated. :smiley:


PhrozenByte:
Please refer to #343 for feedback about this PR. Thank you @smcdougall for this PR! :+1:

nginx.md
```diff
@@ -0,0 +1,156 @@
...
+    _title: URL Rewriting
+    pico-in-document-root: Pico in Document Root
+    pico-in-a-subfolder:
+      _title: Pico in a Subfolder
+      regex-and-nginxs-processing-order: Regex and Nginx's Processing Order
+  advanced-configuration-tips:
+    _title: Advanced Configuration Tips
+    modular-pico-config: Modular Pico Config
+nav-url: /docs/
+gh_release: v1.0.2
```
PhrozenByte: Not used

nginx.md
```diff
@@ -0,0 +1,156 @@
...
+redirect_from: /nginx.html
```
PhrozenByte: The `redirect_from` plugin is used for BC; this is a new page, thus not necessary.

smcdougall:
I wasn't sure about those.  I had just based it on another page, so I left them in.  Thanks.

nginx.md
```diff
@@ -0,0 +1,163 @@
...
+
+There are several ways to approach Pico's url-rewriting in Nginx.  We'll be covering two different approaches in this document, but there really is no right answer.
+
+### Pico in Document Root
+
+If your Pico installation is in the `Document Root` of your website, then configuration is relatively easy.
+
+```
+location / {
+	try_files $uri $uri/ /?$uri&$args
```
PhrozenByte: Missing ';'

nginx.md
```diff
@@ -0,0 +1,163 @@
...
+	if (!-f $document_root$fastcgi_script_name) {
+		return 404;
+	}
+
+	fastcgi_pass unix:/var/run/php5-fpm.sock;
+	fastcgi_index index.php;
+	include fastcgi_params;
+
+	# Inform Pico we will be rewriting URL's for it.
+	fastcgi_param PICO_URL_REWRITING 1;
```
PhrozenByte: Shouldn't this be in the Pico location block?

smcdougall:
Thanks, those semi-colons are the bane of my existence.  I leave them off at least once every time I edit a config file. >.<

Unfortunately, it won't work in the Pico location block.  It needs to send the variable externally to PHP, and because of how the blocks are processed, it won't "remember" it from the Pico block when it gets to the php block.

It's a two-step process, first the url is rewritten to "/?some/page/url" using the Pico block, then it checks it again and matches it against the PHP block.  It then sends the page to `php-fpm` for external processing, which is when it needs to know the variable.

PhrozenByte:
Ah, ic, thanks for the explanation :smiley:

_docs/config.md
```diff
@@ -16,12 +17,17 @@ Pico's default URLs (e.g. http://example.com/pico/?sub/page) already are very us
...
 
-If you're using Nginx, you can use the following configuration to enable URL rewriting. Don't forget to adjust the path (`/pico`; line `1` and `4`) to match your installation directory. You can then enable URL rewriting by setting `$config['rewrite_url'] = true;` in your `config/config.php`.
+#### Nginx Configuration
 
-    location ~ ^/pico(.*) {
-        index index.php;
-        try_files $uri $uri/ /pico/?$1&$args;
-    }
+If you're using Nginx, you can use the following configuration to enable URL rewriting. Don't forget to adjust the path (`/pico`; line `1` and `4`) to match your installation directory. You can then enable URL rewriting by setting `$config['rewrite_url'] = true;` in your `config/config.php`.
+```
```
PhrozenByte: GitHub Pages (i.e. [Kramdown](http://kramdown.gettalong.org/)) unfortunately doesn't support fenced code blocks correctly (three different Markdown parses are a pain...). See http://smcdougall.github.io/Pico/docs/#nginx-configuration. But indentation works fine :smiley:

smcdougall:
Huh, wow, I did *not* notice that.  Why does it work in `nginx.md` then?  Wouldn't they be processed the same?

smcdougall:
Oops, that was supposed to reissue the same commit... guess I did it wrong. :laughing:

PhrozenByte:
Ah, no, Kramdown *does* support fenced code blocks, but it requires you to add a empty line before and after the fenced code block.

smcdougall:
Ah, okay.  Should be fixed now.

_docs/config.md
```diff
@@ -16,12 +17,23 @@ Pico's default URLs (e.g. http://example.com/pico/?sub/page) already are very us
...
 If you're using the Apache web server, URL rewriting should be enabled automatically. If you get an error message from your web server, please make sure to enable the [`mod_rewrite` module][ModRewrite]. Assuming rewritten URLs work, but Pico still shows no rewritten URLs, force URL rewriting by setting `$config['rewrite_url'] = true;` in your `config/config.php`.
 
-If you're using Nginx, you can use the following configuration to enable URL rewriting. Don't forget to adjust the path (`/pico`; line `1` and `4`) to match your installation directory. You can then enable URL rewriting by setting `$config['rewrite_url'] = true;` in your `config/config.php`.
+#### Nginx Configuration
 
-    location ~ ^/pico(.*) {
-        index index.php;
-        try_files $uri $uri/ /pico/?$1&$args;
-    }
+If you're using Nginx, you can use the following configuration to enable URL rewriting.  You'll need to adjust the path (`/pico` on lines `5` and `7`) to match your installation directory. You should also add `fastcgi_param PICO_URL_REWRITING 1;` to your PHP Location Block to enable URL rewriting in Pico.
```
PhrozenByte: For the "short" setup example it may be easier for a user to set `$config['rewrite_url'] = true;` in `config/config.php` instead. From my rather unexperienced point of view regarding nginx's configuration, this would probably end up in trial & error and a angry myself blaming Pico why it isn't working. :smile:

_docs/config.md
```diff
@@ -16,12 +17,23 @@ Pico's default URLs (e.g. http://example.com/pico/?sub/page) already are very us
 
 If you're using the Apache web server, URL rewriting should be enabled automatically. If you get an error message from your web server, please make sure to enable the [`mod_rewrite` module][ModRewrite]. Assuming rewritten URLs work, but Pico still shows no rewritten URLs, force URL rewriting by setting `$config['rewrite_url'] = true;` in your `config/config.php`.
 
-If you're using Nginx, you can use the following configuration to enable URL rewriting. Don't forget to adjust the path (`/pico`; line `1` and `4`) to match your installation directory. You can then enable URL rewriting by setting `$config['rewrite_url'] = true;` in your `config/config.php`.
+#### Nginx Configuration
```
PhrozenByte: Can you please add a "Learn more..." here (see Upgrade section above)? Sure, this is up to change, but in the meantime...

_docs/config.md
```diff
@@ -16,12 +17,23 @@ Pico's default URLs (e.g. http://example.com/pico/?sub/page) already are very us
...
+	return 404;
+}
+
+location ~ ^/pico(.*) {
+	index index.php;
+	try_files $uri $uri/ /pico/?$1&$args;
+}
+```
+
+This configuration should provide the *bare minimum* you need to for Pico.  Nginx is a very extensive subject.  If you have any trouble, please read through our page on [Nginx Configuration][NginxConfig].  Don't be afraid to open a [new issue](https://github.com/picocms/Pico/issues) on GitHub or contact us at [#picocms on Freenode IRC](https://webchat.freenode.net/?channels=%23picocms) for additional assistance.
```
PhrozenByte: Please don't mix inline and footnote URLs (i.e. move the URLs to the bottom)

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+In comparison, Nginx's "Centralized" configuration is all located in one place.  On a Linux server, this configuration would usually be located in `/etc/nginx`, but this may vary by distro and OS.  While [configuration of Nginx](https://www.nginx.com/resources/wiki/start/) as a whole is out of the scope of this document, we hope to provide you with enough information to get over any hurdles you may encounter.
+
+If you are migrating from Apache, [this article](https://www.digitalocean.com/community/tutorials/apache-vs-nginx-practical-considerations) (also linked above) and [this tutorial](https://www.digitalocean.com/community/tutorials/how-to-migrate-from-an-apache-web-server-to-nginx-on-an-ubuntu-vps) provide an excellent overview on the differences between the two and the changes you'll have to make.
+
+## Getting Started
+
+While the [example]({{ site.github.url }}/docs#nginx-configuration) provided on the previous page is a good starting point, it will likely not be enough to fully configure your Pico installation.
```
PhrozenByte: I don't think that this is true, otherwise the provided example would be wrong or seriously incomplete :wink:

The difference rather is that the example provided here is more comprehensive, also covers the relevant general webserver configuration, explains the *Why* and weighs alternatives. :smiley:

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+There are three main parts to configuring a Pico site under Nginx, in addition to general server configuration.  The three sets of rules we will be developing provide the following functions: Denying access to Pico's internal files, configuring PHP, and setting up Pico's url-rewriting.  Although it's arguably the most important function, we'll be configuring url-rewriting last due to the order that Nginx processes its config.  We'll discuss that in more detail when we get to it below.
```
PhrozenByte: I'm no native speaker, but "three main parts to configuring" sounds a bit odd. :unsure:

PhrozenByte: Hmm... The "three sets of rules we will be developing" (btw: you're also covering the general webserver config, so shouldn't it be four?) just screams to use sub-sections, i.e.
* Getting started
  * General Server Configuration
  * Denying Access to Pico’s Internal Files
  * PHP Configuration
  * URL Rewriting
* Example Configuration
* Advanced Configuration Tips

smcdougall:
Yeah, that was written with the original example in mind.  What do you think it should say?

"While the example... starting point, here we will provide a more in-depth look at Nginx configuration."

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+## General Server Configuration
+
+Nginx's server configuration is broken into sections called `blocks`.  Each website has its own `server` block inside your Nginx config.  While configuration of Nginx sites, virtual hosts, and other aspects of this topic are outside the scope of this guide, we'll provide enough to at least get you started with Pico.
```
PhrozenByte: I don't think that `blocks` is a actual "name". Just "Nginx's server configuration is broken into sections." should be fine :smiley:

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+	listen 80;
+
+	server_name www.example.com example.com;
+	root /var/www/example;
+
+	index index.php;
+}
+```
+
+Let's break down this example.  The first line, `listen 80` tells Nginx to listen on port 80 for incoming connections.  This is the default port used by web traffic, and what you'll want to use 99% of the time.  `server_name` is where you specify what domain name or names match this configuration.  You'll likely want to include your domain both with and without the `www.` subdomain.  `root` lets you specify the Document Root for this site.  This is usually going to be where you've installed Pico, but ultimately it depends on your configuration.  `index index.php` tells Nginx that your site's index file will be called `index.php`.  This is necessary for Pico, however you can use something like `index index.html index.htm index.php` if you'd like Nginx to also search for a non-Pico html index file.
```
PhrozenByte: Can you please add a URL to a tutorial how to configure HTTPS with Nginx (plus a sentence that SSL isn't covered by this example, but its use is recommended)? Nowadays HTTPS is what you should want to use 99% of the time :smile:

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+Below we'll add a few more sections to this server configuration.  All of the following examples should be placed **inside** the server block, before the closing `}`
+
+## Denying Access to Pico's Internal Files
+
+One of the important features provided by Pico's `.htaccess` is the denial of access to some of Pico's directories.  There are two simple reasons for this, added security, and ease-of-use.  We don't want anyone snooping around and reading files they shouldn't be (like our Pico `config.php`), but we also don't want people accidentally navigating to a raw markdown file or viewing directory indexes (not actually an issue, since they're disabled unless specified in Nginx, but this isn't always the case in Apache).
```
PhrozenByte: > since they're disabled unless specified in Nginx, but this isn't always the case in Apache

This is also the case in Apache, but it's a common misconfiguration. I'm not sure whether this is true for nginx, too.

smcdougall:
>  you're also covering the general webserver config, so shouldn't it be four?

I don't know, I counted it separately since it wasn't something specific to Pico...

> * Getting started
>    * General Server Configuration
>    * Denying Access to Pico’s Internal Files
>    * PHP Configuration
>    * URL Rewriting

I used "Getting Started" as an introduction to the tutorial portion of the page.  Before that, it was more informative than instructive.  It was just meant to be a transition, so I'm not sure if all the other steps should be part of it.

If you're configuring the URL Rewriting, you aren't really "getting started" anymore.

It would work okay as a subsection though, but maybe called something other than "Getting Started".  "Nginx Configuration" for example, but something that isn't also the title of the page.

smcdougall:
> I don't think that blocks is a actual "name". Just "Nginx's server configuration is broken into sections." should be fine :smiley:

https://www.nginx.com/resources/wiki/start/topics/examples/server_blocks/

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+You should always use the following rule in the Nginx config of your Pico site:
+
+```
+location ~ /(\.htaccess|\.git|config|content|content-sample|lib|vendor) {
```
PhrozenByte: This rule also affects non-Pico subdirectories, isn't it? Especially blocking access to `content` globally may cause trouble with other software. Does Nginx allow cascading `location` blocks? Otherwise a users needs to adjust the base path (i.e. `/` or `/pico/`) manually :unamused:

In general: Shouldn't this section come after the PHP configuration or is this necessary because of Nginx's processing order? The document should follow a golden thread from "General Configuration" to "Pico's configuration" - at least as long as Nginx's configuration order doesn't foil it.

smcdougall:
> Can you please add a URL to a tutorial how to configure HTTPS with Nginx (plus a sentence that SSL isn't covered by this example, but its use is recommended)? Nowadays HTTPS is what you should want to use 99% of the time :smile:

Good idea, I'll probably link to [Let's Encrypt](https://letsencrypt.org/getting-started/).  I've tried it, and it's really easy to use.  Much more so than other solutions I've tried.

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+This rule returns a 404 (file not found) error if someone tries to access `.htaccess` or Pico's other internal files.  Remember, `.htaccess` is a configuration file for Apache, and Apache would have denied access to it automatically.  While it doesn't do anything under Nginx, you may have unknowingly left it in your Pico directory, as it is hidden by default.  While you can safely remove `.htaccess` from your Pico installation, we'd recommend you keep it in this rule anyway just in case.  For example, you may upgrade Pico in the future and forget to remove it!  In general though, this rule is for security, and a similar rule is usually recommend in Nginx, even outside of running Pico.
+
+The rest of the rule returns a 404 page if the user tries to navigate into Pico's internal file structure.  We recommend this as it's generally a good practice.  Users's don't need access to these files, so why allow it?
+
+## PHP Configuration
+
+This is a topic outside the realm of this document.  Unlike Apache (which sends every document to PHP automatically), Nginx is more streamlined.  It needs to be *told* to send a file to an external PHP processor.  It does not handle PHP itself.
```
PhrozenByte: > Unlike Apache (which sends every document to PHP automatically), Nginx is more streamlined.

That's not entirely true. :wink: Actually Apache and Nginx handle PHP the exact same way: you configure a file extension (or multiple file extensions) and the webserver sends all requests to files with this file extension to PHP (still abridged for Apache; Apache uses MIME types, but they ultimately also mostly depend on file extensions). The difference between `mod_php` (Apache) and FastCGI (both Apache and Nginx) is about how the webserver *communicates* with PHP (i.e. how PHP processes are handled, so it's a more technical question), not how requests are handled. The difference between Apache and Nginx is that Linux distributions ship more intelligent default configurations for Apache, so Apache users don't have to deal with this on their own.

smcdougall:
> > since they're disabled unless specified in Nginx, but this isn't always the case in Apache

> This is also the case in Apache, but it's a common misconfiguration. I'm not sure whether this is true for nginx, too.

I'm not sure.  I had too look up the configuration of Indexes when I was writing that part.  I think this is what you have to use: http://nginx.org/en/docs/http/ngx_http_autoindex_module.html

I had to search for it, and hadn't come across it *at all* beforehand.  I've never seen any examples specify "autoindex off" in the same way that many `.htaccess` files do.

So yeah, I'm not sure, but I haven't really seen it mentioned anywhere as an issue.

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+Configuring PHP is a topic that is largely up to the OS you are using.  The examples I'm going to provide here apply to Ubuntu 14.04, but they also require external configuration of `php-fpm` or another php processor.
```
PhrozenByte: Your example isn't OS-Dependant. Apart from file paths (i.e. where the FastCGI socket is located, more precisely `/var/run/php5-fpm.sock`) you can use the same config for Windows :smiley:

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+Your PHP configuration will look something like this:
+
+```
+location ~ [^/]\.php(/|$) {
+	fastcgi_split_path_info ^(.+?\.php)(/.*)$;
+
+	# Protection Against "cgi.fix_pathinfo = 1"
```
PhrozenByte: Shouldn't this be `try_files` as discussed in https://github.com/picocms/Pico/issues/343#issuecomment-213793577?

smcdougall:
> This rule also affects non-Pico subdirectories, isn't it? Especially blocking access to content globally may cause trouble with other software. Does Nginx allow cascading location blocks? Otherwise a users needs to adjust the base path (i.e. / or /pico/) manually :unamused:

Yes, it blocks access to them anywhere.  I thought it'd be better than having to customize it depending on your subfolder.  I found the subfolder issue a little annoying, so I tried to condense it all into the Rewrite section.  Having "Pico in Document Root" and "Pico in a subfolder" for every section sounded too confusing.  I've been trying to avoid differing configurations wherever possible.

Nginx does allow nesting location blocks though, so there is some potential there.

Keep in mind though, this configuration only affects a given virtual host.  But, I guess I could see the issue if they were for example, running Pico in `/pico/` and a forum in `/forum/`, `/forum/content/` would then be blocked.

*sigh* This is the trouble with coming up with a *single* example that works in all cases.

> In general: Shouldn't this section come after the PHP configuration or is this necessary because of Nginx's processing order? The document should follow a golden thread from "General Configuration" to "Pico's configuration" - at least as long as Nginx's configuration order doesn't foil it.

It needs to come before PHP.  If it comes after PHP, then Nginx will match that rule first, meaning any PHP files (`config.php` for example) won't get denied.

smcdougall:
> The difference between Apache and Nginx is that Linux distributions ship more intelligent default configurations for Apache, so Apache users don't have to deal with this on their own.

Ouch.  lol.

How would you have me explain the differences then?  It's still somewhat internal vs external, as `mod_php` is part of Apache, while `php-fpm` is a separate software from Nginx.

I feel like most of your issue would be addressed by removing `is more streamlined.  It` and combining the two sentences.  Maybe also drop `It does not handle PHP itself.`, to minimize the comparison between the two.

smcdougall:
> Your example isn't OS-Dependant. Apart from file paths (i.e. where the FastCGI socket is located, more precisely /var/run/php5-fpm.sock) you can use the same config for Windows :smiley:

Good to know, but I mainly wanted a reason why a user shouldn't copy and paste it (and also, I haven't tested it on anything but my own server).  Should I just say that configuration "differs slightly" depending on your OS?

smcdougall:
> Shouldn't this be try_files as discussed in #343 (comment)?

I never got a chance to test that and I guess it got lost in the discussion afterward.  If you *know* it to be working, I'll replace the `if` statement with that.  Otherwise I want to test it first and make sure it protects against the security issue.

smcdougall:
Yay, I've caught up with you. :wink:

This comment thread is just about as out-of-order as it could be.

smcdougall:
> I'll probably link to Let's Encrypt.

Have you tried Let's Encrypt before?  I believe it's fairly new.  Just wondering if it should be a casual recommendation or a project-backed one.  Probably something like "You can check out" vs "We recommend."

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+The line `fastcgi_param PICO_URL_REWRITING 1;` informs Pico that we will be rewriting url's in Nginx.  This prevents Pico from using its own url-rewriting, and gives us nicer url's.
+
+## URL Rewriting
+
+There are several ways to approach Pico's url-rewriting in Nginx.  We'll be covering two different approaches in this document, but there really is no right answer.
+
+### Pico in Document Root
+
+If your Pico installation is in the `Document Root` of your website, then configuration is relatively easy.
```
PhrozenByte: Why is `Document Root` formatted as `code`?

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+```
+location / {
+	try_files $uri $uri/ /?$uri&$args;
+}
+```
+
+This rule tells Nginx that whenever it looks up a url within your site, it should first look for a file of that name, then a directory, and finally, if neither exist, it should rewrite the url for Pico.  When the url is rewritten for Pico, it is formatted as `/` followed by a query string (starts with `?`) containing the file path (`$uri`) and then followed by any other query arguments (`$args`).  By referencing `/`, Nginx will load `index.php` (as defined in our server config above) and pass the rendering along to your php processor.
```
PhrozenByte: > By referencing `/`, Nginx will load `index.php`

Reading this, we can save a round trip by directly specifying `/index.php?$uri&$args`.

smcdougall:
> Why is Document Root formatted as code?

Sometimes I like how it formats things.... :unamused:

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Pico in a Subfolder
+
+If your Pico installation is in a subfolder, configuration is slightly more complicated, but should be easy if you follow along.
+
+```
+location ~ ^/pico(.*) {
+	try_files $uri $uri/ /pico/?$1&$args;
```
PhrozenByte: `/pico/index.php?$1&$args;`

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+}
+```
+
+You'll notice that similar to our last example, we're sending Nginx to `index.php` with a query string of the url, but this time in a subfolder named `pico`.  What's different is that we're using `$1` to reference the page url instead of `$uri`.  This is because `$uri` will contain the entire url from the Document Root, but we only want the part that comes *after* the `/pico` subfolder.  Since we can't use `$uri` for this, we're using a Regular Expression (also called "regex") to determine the url for us.  This `location` rule looks for any url that starts with `/pico`, and takes note of whatever comes afterward.  In your own configuration, you'll need to replace `pico` on both lines with the location of your own subfolder.
```
PhrozenByte: "What's different is...", "What is different is..." - shouldn't this be something like "The difference is..."?

smcdougall:
> Reading this, we can save a round trip by directly rewriting to /index.php?$uri&$args.

I don't know if it really saves a "round trip" or not.  My assumption for how it works is that it would just "tack on" the name of the index file if unspecified.

Either way, I'll add `index.php` to match it up closer with `.htaccess`.

nginx.md
```diff
@@ -0,0 +1,190 @@
...
+}
+```
+
+Again, please note that this is only provided as an **example**.  You should not copy it, but only use it as a reference.  Your own configuration will depend very much on your own system configuration.  If you do copy this code, be sure to modify it according to your own requirements.
+
+## Advanced Configuration Tips
+
+### Modular Pico Config
+
+Let's say you're a real Pico enthusiast and have several Pico websites running on the same server.  You may get tired of writing all these rules into each and every server configuration.  An easier solution might be to place all the common components (index, access denials, php rules) into a separate file and include it using `include /absolute/path/to/file`.  You could also add the rewrite rule to this file, but a better option would be to include a second file, that way you can chose to include it *only* when Pico is located in your Document Root.
```
PhrozenByte: I think it's a good idea to let the components, i.e.

> An easier solution might be to place all the common components (index, access denials, php rules) into a separate file

directly link to the appropriate sections above :smiley:

smcdougall:
Considering you're from Germany.... it would probably be in poor taste to call you a "Documentation Nazi", huh? :unamused:

In case that's too lost in translation, It was a play on the term "[Grammer Nazi](http://www.urbandictionary.com/define.php?term=Grammar+Nazi)" more than anything.

Seriously, I appreciate all the help. :wink:

...debated way too long before sending this comment :unamused:

smcdougall:
> I think it's a good idea to let the components directly link to the appropriate sections above :smiley:

:+1:

smcdougall:
I've tested `test try_files $uri =404;` and it seems to work properly.  Oddly enough, even with `cgi.fix_pathinfo = 1`, my server returned an "Access Denied" message from PHP (it was not a message from Nginx).  There must be another security option somewhere else in Ubuntu's default config that I don't know about.

smcdougall:
:unamused: I just realized I could have replied to each of your comments *inside* the line-note.

I don't know,  I guess I just thought that button was for adding a *new* note. :disappointed:

PhrozenByte:
Okay, I've finished reading. Reading and fixing should never happen simultaneous :wink: You can btw directly answer to my comments by clicking the `Add a line note` button, otherwise the comments get completely out of order and it's extremely hard to follow the discussion. Anyway, I'll answer your comments here this time.

bildschirmfoto von 2016-04-29 16 55 06

**General notes:**
I love it! :heart: Sure, there are small issues, but none of them is even close to being "mission critical" :smiley: As you've [already noticed (:smile:)](https://github.com/picocms/Pico/pull/345#issuecomment-215741169) I'm a perfectionist (:wink:), so please don't be offended by so many comments - they are all just very minor issues I stumbled over. I could have fixed them after merge myself, but it's your document, so I thought it would be better to tell you where I see open issues instead.

After you've finished your work I'll read it again and merge it. Again, I cannot emphasise enough that the document is great! :+1:

**Section "Getting Started", first paragraph:** https://github.com/picocms/Pico/pull/345#issuecomment-215705468
> While the example... starting point, here we will provide a more in-depth look at Nginx configuration.

Yeah, I like that! :+1:

**Splitting the document into subsections:** https://github.com/picocms/Pico/pull/345#issuecomment-215708293
> I used "Getting Started" as an introduction to the tutorial portion of the page.

True. "Nginx Configuration" sounds good. This is completely up to you, it just came to my mind immediately after reading this sentence.

**How to name `blocks`:** https://github.com/picocms/Pico/pull/345#issuecomment-215708912
Then maybe simply just "Nginx's server configuration is broken into blocks." or something like that? I just stumbled over it while reading :smile:

**Nginx and HTTPS:** https://github.com/picocms/Pico/pull/345#issuecomment-215709592 + https://github.com/picocms/Pico/pull/345#issuecomment-215731584
Yeah, I'm using Let's Encrypt, too (and I love it). 353a17bbc3f623e4728b92dd6810c54d75065e6f is a absolute perfect solution! :+1:

**Directory listing (`autoindex off;` and `Options -Indexes`):** https://github.com/picocms/Pico/pull/345#issuecomment-215711817
Ah, I see. Yeah, then it's probably best to simply remove the half sentence in brackets as it doesn't apply to nginx anyway.

**Denying access globally:** https://github.com/picocms/Pico/pull/345#issuecomment-215719269
I look forward to your solution for it :smiley:

**Moving the "Deny access" section below "PHP Configuration":** https://github.com/picocms/Pico/pull/345#issuecomment-215719269
Ah, okay, thanks for clarifying this!

**PHP with nginx vs. PHP with Apache:** https://github.com/picocms/Pico/pull/345#issuecomment-215724461
Yep, simply remove the comparison, it's just fine then :smiley:

**OS-Dependant examples:** https://github.com/picocms/Pico/pull/345#issuecomment-215725785
Hmm... I don't have a real solution for this. But you had one :wink: 353a17bbc3f623e4728b92dd6810c54d75065e6f is just fine!

**`try_files` instead of the very complex `location` regex, `fastcgi_split_path_info` and `if`:** https://github.com/picocms/Pico/pull/345#issuecomment-215726643 + https://github.com/picocms/Pico/pull/345#issuecomment-215756031
Yes, I've tested it, works like a charm. :smiley:

nginx.md
```diff
@@ -0,0 +1,192 @@
+---
```
PhrozenByte: Can you please move the file to `in-depth/nginx.md`?

smcdougall:
> [That line you removed about answering me *this one time*]

My excuse is that I was thinking that adding a new line note would put it at the end of the thread anyway.  I didn't realize until I clicked on it that it would create a little sub-thread for the reply.  It makes sense though, and seems handy.  I'll use that next time.

>  I'm a perfectionist (:wink:)

Me too.  I understand. :laughing:

smcdougall:
> try_files instead of the very complex `location` regex, `fastcgi_split_path_info` and `if`

So it's safe to remove `fastcgi_split_path_info`?  I haven't tried that myself, but the point of the line seems to be to separate out any path following the `.php` file (`example.com/index.php/some/url`) and give it to PHP as a variable.  Since Pico instead uses a query string internally, I can see why that wouldn't be needed.  Just wanted to make sure there's no other purpose it might server for us.

Honestly, I've just been including the line because I've seen it a lot in examples... I don't actually know what purpose it serves on the PHP side of things or what that path information is used for. :sweat_smile:

PhrozenByte:
Yes, you can safely remove `fastcgi_split_path_info`. You can also use a much simpler regex like `location ~ \.php$ { }`.

The purpose of this directive is to imitate the `PATH_INFO` feature to allow SEO-like URLs out-of-the-box. `PATH_INFO` actually is part of the official CGI specification (see https://tools.ietf.org/html/rfc3875#section-4.1.5), but is practically limited to Apache (see https://httpd.apache.org/docs/2.2/mod/core.html#acceptpathinfo). Alternative webservers (like nginx - or IIS, lighttpd, httpd...)support `PATH_INFO` with workarounds only. Many configurations out there don't have those workarounds enabled. Thus it's usually a bad idea to use `PATH_INFO`. However, there are (better) alternatives - like `QUERY_STRING`.

smcdougall:
*sigh* So I don't really have a better option for the denial rule at the moment.  I added a little note about it in-case it causes issues.  I don't really think it's worth the extra effort at this point when I'll probably have to rewrite half the document for `1.1` anyway.

In the future, a solution might be a `location` block containing *all* the Pico rules.  Using a setup like that, there should be a way to state Pico's location *once* and have all the other rules conform to it.

I experimented a little while back with adding a separate `root` declaration inside of a `location` block and it did successfully change the `$document_root` variable *inside* the location block.  The problem was that it didn't impact `$uri` at all.  I think this was due to the change being after Nginx had created the $uri (it was inside the rewrite rule block).

Perhaps by declaring a Pico `location` with a new `root`, we could treat all the nested locations (the existing blocks) as if Pico were installed in Document Root.  If this worked, we could potentially provide *one* Nginx example that would work no matter where Pico was installed.  You'd enter Pico's location once and all subsequent rules would use it.  This would also make all our rules affect *only* Pico, and eliminate the need to worry about interfering other software on the same virtual host.

Just speculating at the point, but it would be worth looking into in the future. :smile:

---

Also, in regards to adding a new subsection to the document where `Getting Started` is...  I'm not sure it's necessary.  I'm just looking at the rendered page and imagining adding a new header.  It feels a little redundant.  There would only be two sections not inside it (Example Config and Advanced Tips), so it feels a bit like the entire document would be inside the new section.  The document as a whole is about "Nginx Configuration", and every section in it is directly related to that topic (except for where I've already broken it down).

I don't know.  As always, I'm open to any arguments you have in favor of it, I just don't think it adds any benefit right now.

---

And, I think that's everything.  Let me know if I've forgotten to address any points we discussed.

And sorry for the extra long reply. :sweat_smile:

PhrozenByte:
Using `root` sounds promising, I look forward to your changes for Pico 1.1. I'm fine with not adding a superior section, it was just an idea. Maybe when you update the document for Pico 1.1. We'll see :smiley:

I've merged your PR with some very minor changes (see fa07158a032b22cdc841e0ff0208f58096c57c93). :tada: Thank you for your efforts! :+1: :smiley:

PR #346, opened 2016-04-21
Use http or https depending on page access

ghuron:


PhrozenByte:
Thank you @ghuron! :+1:

smcdougall:
I actually have a question about this.  Why not *always* use https for the connection to Google Fonts?  That's what I've done in NotePaper and it works fine.  Even if the regular page is loaded over http, that doesn't mean the connection to Google Fonts has to use http as well.

It just seems like being secure by default would be the better option. :confused:

PhrozenByte:
Good point, thank you @smcdougall :smiley:

PR #359, opened 2016-06-12
Documentation Rewrite

smcdougall:
Thought of making this a new Issue, but since I've made some commits already (notes and minor fixes), I decided to just PR, as it'll need to be one eventually.

Currently, this contains some unmerged commits from #352.

New commits start at `Added Improvement Notes to Docs. Minor fixes and improvements.` https://github.com/picocms/Pico/commit/3a76be4356ad418855ee5c320728b75e58448ad2

The idea here is going to be a full (or close to) rewrite of the existing Documentation into smaller, more in-depth pages.

Since the plan is to redesign the docs around this new style, it'll only really be merge-able when _all_ the work is done.  For this reason I've decided to make it one PR and not do a PR per-page.  If this changes, or I write a decent, standalone page, we can always merge them in on a case-by-case basis.

Discussion on this topic was previously started on #352.

Here's some quotes from that thread:
- Page Topics

> - An "Installing Pico" page that covers, in-depth, the concepts of installing and configuring Pico, your webserver (well, links to external resources anyway) and etc. Everything up until you're looking at Pico's sample content.
> - A "Creating Content" page that goes over YAML and Markdown.
>   Perhaps a general, "Usage" type page that covers things like how URLs work, how to lay out and organize your website, separating content and assets, more about config options, etc.
> - Customization: An in-depth guide to installing community themes, creating your own, understanding Twig, Pico variables, everything. Not plugins though.
> - A separate page for Plugins (and plugin development).
> - Maybe a combination of "Getting Help" and "Contributing" (or "Getting Involved"). It should be a generally community-focused page. On that note, I think it would be great if we eventually had a place for non-development, non-bug related community discussions. I don't want to say a "forum" because that feels like a really outdated notion, but a "place" to develop a community nonetheless.
- Table of Contents

> When on the Overview Page:
> 
> ```
> * Installing Pico
> * Creating Content
> * Customization
> * Getting Help
> ```
> 
> When on Creating Content Page:
> 
> ```
> * Installing Pico
> * Creating Content
>    * YAML
>    * Markdown
>    * Etc
> * Customization
> * Getting Help
> ```

---
- Advanced asset handling guide https://github.com/picocms/Pico/pull/368#issuecomment-243415712
- Definition Lists for Twig Terms https://github.com/picocms/Pico/pull/358#discussion_r83863940
## Cookbook Examples
- ~~https://github.com/picocms/Pico/pull/334#issuecomment-255223633~~

PR #368, opened 2016-08-29
Add support for serving static files from content folder

mpavelka:
Blog posts usually contain images. Adding this code makes it convenient for bloggers to include them in Pico content.
In a ```content/content-subfolder/post.md``` file add pictures like this:
```
![My image](my_image.jpg)
```
or like this without mod rewrite enabled
```
![My image](?content-subfolder/my_image.jpg)
```

### DIFF
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -124,6 +124,14 @@ class Pico
     protected $requestFile;
 
     /**
+     * Absolute path to the content file being served
+     *
+     * @see Pico::getRequestFile()
+     * @var boolean|false
+     */
+    protected $requestFileIsStatic;
+
+    /**
      * Raw, not yet parsed contents to serve
      *
      * @see Pico::getRawContent()
@@ -299,6 +307,16 @@ class Pico
 
         // load raw file content
         $this->triggerEvent('onContentLoading', array(&$this->requestFile));
+
+        if ($this->getRequestFileIsStatic()) {
+            // discover mime type
+            $mimeType = mime_content_type($this->requestFile);
+            $fp = fopen($this->requestFile, 'rb');
+            header("Content-Type: ".$mimeType);
+            header("Content-Length: " . filesize($this->requestFile));
+            fpassthru($fp);
+            exit;
+        }
 
         $notFoundFile = '404' . $this->getConfig('content_ext');
         if (file_exists($this->requestFile) && (basename($this->requestFile) !== $notFoundFile)) {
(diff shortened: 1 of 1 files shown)


PhrozenByte:
First of all, thank you for your PR and making Pico even better @mpavelka! :+1:

Unfortunately I've to tell you that this is against Pico's recommended way of managing assets. Let me explain the reasons:

Passing static files (like images) through PHP is a highly unperformant solution. A user usually generates just a single dynamic/interpreted request per page impression (i.e. the actual page), plus a much higher number of static requests (e.g. stylesheets, images, JavaScript ...). A dynamic/interpreted request causes an order of magnitude higher performance impact on your server than a static request and is - by far! - the most important limiting factor of how many concurrent requests your webserver can handle.

By serving just one image through PHP, a single page impression results in two dynamic/interpreted requests. Serving just two or three images are then very likely sufficient to cause the same performance impact as two "normal" page impressions. This bisects the number of concurrent requests your webserver can handle!

Therefore passing static file requests through PHP is no option. However, I agree that users want to store their assets together with their contents. Nevertheless, we deny access to the `content/` folder for security reasons. This protection actually isn't necessary for a vanilla Pico installation because it serves all contents of the `content/` folder anyway, however, plugins may restrict access to certain pages. As this may cause hidden security issues for users who want to use such plugins, we deny access to the `content/` folder by default ("security by default").

This unfortunately leads to a conflict of aims between convenience and security. The obvious solution is to still deny access by default, but allow access to certain files (e.g. by matching MIME types). However, configuring your webserver to do so is (besides its still not negligible performance impact) a pretty complex topic and varies from webserver to webserver. I think that this is a topic for very advanced users only.

Thus we recommend to create a distinct `assets/` folder. This definitely is not a convenient solution, but - as far as I know - the only practicable.

Personally I recommend you to configure your webserver to behave as elucidated above. If you still want to pass asset files through PHP, I recommend you to create a plugin (simply hook into Pico's `onRequestFile` event). Again, thank you for contributing! :+1:

mpavelka:
@PhrozenByte, thank you very much for the most exhaustive explanation to the topic, it makes a lot of sense.
I'm sure I'm not the first one to wonder about this problem, perhaps it would help others too to include a little how-to in docs/about page.

PhrozenByte:
@mpavelka, you're welcome, I've added this to Pico's ToDo list to create a appropriate "advanced asset handling howto" :smiley:

@smcdougall, do you want to incorporate the above explanation into #359 (beginning with "Nevertheless, we deny access to the `content/` folder for security reasons. […]")?

smcdougall:
Sure.  :+1:

I've made a note on #359 to come back here and reference your explanation.

PR #384, opened 2017-05-14
CONTRIBUTING.md: Create "Labelling of Issues & Pull-Requests" section (2)

PhrozenByte:
Proposed solution to #308 as an alternative to #310 

Feedback is highly appreciated!

Closes #310, #308


PhrozenByte:
> My only concern would be that this makes the contributing page run a bit long

What do you think about moving everything except for the "Issues" and "Contributing code" sections to a separate `DEVELOPMENT.md`? These sections aren't really related to "contributing", they rather explain in-depth how we want to organize our project...

smcdougall:
No, I think having everything on one page is the right approach for now.  It's a good all-in-one place for "the rules".  Until it grows a little bigger, I don't think it's outgrown `CONTRIBUTING.md` yet.

Development doesn't sound like the right name either though.  Obviously, we should call it `HOW-WE-ORGANIZE-OUR-PROJECT-IN-DEPTH.md`.  That'd be a much better name.

PR #389, opened 2017-06-15
Enhancement: added %include% tag

AlexanderKrutov:
Introduced `%include%` tag that allows to include file content from an external file.

Usage in content files:

```md
%include% "/path/to/file/to/include.md"
```

Included file can have meta-variables to be replaced with values defined in parent file.
For example:

**parent.md**:
```md
---
username: John Doe 
---
Hello, %include% "child.md"
```

**child.md**:
```md
%meta.username%. This is an example how include tag works.
```

Will be processed to:
```md
Hello, John Doe. This is an example how include tag works.
```

### DIFF
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -875,7 +875,7 @@ class Pico
             . "(?:(.*?)(?:\r)?\n)?(?(2)\*\/|---)[[:blank:]]*(?:(?:\r)?\n|$)/s";
         $content = preg_replace($metaHeaderPattern, '', $rawContent, 1);
 
-		// process included files
+        // process included files
 		if (preg_match('|%include%\s*(["\'])(.*?)\1|smi', $content, $matches) > 0) {
             $includeFile = $this->getConfig('content_dir') . $matches[2];
             if (!file_exists($includeFile))
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -876,14 +876,16 @@ class Pico
         $content = preg_replace($metaHeaderPattern, '', $rawContent, 1);
 
         // process included files
-        if (preg_match('|%include%\s*(["\'])(.*?)\1|smi', $content, $matches) > 0) {
-            $includeFile = $this->getConfig('content_dir') . $matches[2];
+        $includeTokensCount = preg_match_all('|%include%\s*(["\'])(.*?)\1|smi', $content, $matches);
+        for ($i = 0; $i < $includeTokensCount; $i++) {
+            $includeFile = $this->getConfig('content_dir') . $matches[2][$i];
             if (!file_exists($includeFile)) {
-                throw new RuntimeException('Required "' . $includeFile . '" not found');
+                throw new RuntimeException('Unable to include file "' . $includeFile .
+                    '" referenced from "' . $this->requestFile . '"');
             }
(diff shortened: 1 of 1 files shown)


PhrozenByte:
First of all: Thank you for your contribution! :+1:

Unfortunately there are some problems with your PR:

* The file to include isn't checked in any way, just think of `%include% "../../../../../etc/passwd"`. Starting with Pico 2.0 (see #334) you can use the `Pico::resolveFilePath()` method for things like that.
* The exception message should include the name of the file which is trying to include the missing file.
* It should be possible to include multiple files, not just a single one. However, be careful about recursion (what might be a valid feature, but circular includes are bad).

However, I'm not 100% sure whether we should support a `%include%` tag in Pico's core. We try to follow a pretty strict separation of content and presentation, i.e. you should think of Markdown contents more like a (pretty powerful) "input field" in some CMS's admin panel (like WordPress). All the crazy stuff can (and should) be implemented using Twig in your theme (and we encourage people to create own or at least modify existing themes).

Just an example: A single page of my company's website consists of a bunch of content files in a directory, which are arranged in a certain way (more specifically, they are arranged as "cards" one after another). Your `%include%` tag might be an alternative to this, but it's way less powerful than editing the theme you use.

Just to make this clear, I like your idea and think that there are users that would ❤️ to use a `%include%` tag in their Markdown files, I'm just not sure whether we should include this in Pico's core. However, I didn't make any decision yet, **more input is welcome!**

If we make the decision not to merge this, you have my strong vote to create a plugin and promote it [on our website](http://picocms.org/plugins/). Simply hook into the `onContentParsing` event and do your replacements there, virtually no changes to your code necessary. 😃

```php
class PicoMarkdownIncludesPlugin extends AbstractPicoPlugin
{
    public function onContentParsing(&$rawContent)
    {
        // your code
    }
}
```

AlexanderKrutov:
Hi Daniel,
thank you for the quick reply!

I slightly modified the code according with your comments.
Now it's possible to include namy files not just one (I've missed this case) and exception message is more informative. As for checking of include file path - I've left it as is for now because there is no `Pico::resolveFilePath()` method in the target branch yet. Of course the checking should be added after merging with 2.0.

About adding `%include%` tag in Pico's core - after some thinking over I agree with you that it will be better to make a plugin instead. In other case, there can be cases when using something like `%include%` is prefered. 
For example, if we have some static text/html/md block which used in many content files, like this:
```md
... some page-specific content ...
common_block
... rest page-specific content ...
```

The alternative, as you correctly noted, is use of twig template, but it will require separation of each content file into own "header" and "footer" and single "common_block". That's why I decided to do the PR.

PhrozenByte:
> About adding `%include%` tag in Pico's core - after some thinking over I agree with you that it will be better to make a plugin instead.

Since there was no other feedback and we agree on this, we should probably create a plugin instead. So, after you've created said plugin (you can use e.g. https://github.com/PhrozenByte/pico-file-prefixes as "template" for the repo), I'm happy to officially promote your newly created plugin on [Pico's website](http://picocms.org/plugins/) 😃 👍

PhrozenByte:
I'm closing this for now, waiting for your plugin to promote it on [our website](http://picocms.org/plugins/) 😃 👍

PR #429, opened 2018-04-29
stop config.yml on loading again

JHeimbach:
When loading all other .yml files in the config dir, the config file is loaded again because the if condition is not correct. `$configFile` include the path to the file.


PhrozenByte:
Thanks! :+1:

PR #452, opened 2018-08-06
add support for windows path on evaluateRequestUrl() method

bitmarte:
On evaluateRequestUrl() method we need to consider the windows way to describe a path. In fact it uses slash instead of back-slash.
With this fix we can use PicoCMS on both platform.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,10 +17,10 @@ Pico Changelog
           are only possible with a new major version.
 
 ### Version 2.0.2
-Released: 2018-08-06
+Released: -
 
 ```
-* [Changed] Support Windows path ("\" instead of "/") on evaluateRequestUrl() method
+* [Fixed] Support Windows paths (`\` instead of `/`) in `Pico::evaluateRequestUrl()`
 ```
 
 ### Version 2.0.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ Pico Changelog
           `PicoDeprecated`'s changelog. Please note that BC-breaking changes
           are only possible with a new major version.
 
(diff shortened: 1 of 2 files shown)


PhrozenByte:
Ok, wow, PHP is sometimes really weird... I didn't really understand this patch so I investigated a bit, and it's true: If `$_SERVER['SCRIPT_NAME'] = "/index.php";` (and it looks like this is the *only* situation), `dirname($_SERVER['SCRIPT_NAME']);` returns `\` under Windows. Who knows why, but e.g. `/pico/index.php` results in `/pico`. Weird...

Can you please remove the `CHANGELOG.md` changes or update it to the following?

``````
### Version 2.0.2
Released: -

```
* [Fixed] Support Windows paths (`\` instead of `/`) in `Pico::evaluateRequestUrl()`
```
``````

PhrozenByte:
Great, thank you for your contribution! :+1:

bitmarte:
Hi @PhrozenByte ,
you need to publish a new release to _packagist.org_ in order to include our fix.
In fact when I try to install pico from composer I retrieve the previous version, so I must to override the source from my vendor dist folder.

PhrozenByte:
@bitmarte: Sorry, it took some time, unfortunately my time is very limited at the moment. Here we go: https://github.com/picocms/Pico/releases/tag/v2.0.2

For some reason all Travis builds fail, Travis seems to have massive network problems right now (sometimes resulting in network-related errors, sometimes in stalling and time outs). :unamused: I've tried it >10 times over roughtly 3 hours, but it still doesn't work. Thus there's no pre-built release package yet. I'll try again in a couple of hours... composer-based installations aren't affected by this.

Thanks again! :+1:

PR #457, opened 2018-10-19
"patch-1": Make index.php.dist runnable from top directory.

jjtmp:
Objective is that you don't have to unpack the picocms package in the http(s) servers root path as single files but put it in a directory called "picocms" while keeping it accessible by "http://top.level.domain/?sub".

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,8 @@ Do you know the feeling: You want to install a new website, so you upload all fi
 
 ###### Step 1
 
-[Download the latest Pico release][LatestRelease] and upload all files to the `httpdocs` directory (e.g. `/var/www/html`) of your server.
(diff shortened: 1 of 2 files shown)


PhrozenByte:
Hmm... Unfortunately I don't really understand what you're trying to achieve and why you're trying to achieve it. :confused:

I first though you're trying to move some files to a folder different than `httpdocs`, but this doesn't seem to be the case. What are the advantages of moving Pico to a sub-folder, but not the `index.php`?

jjtmp:
The main goal is to keep the main http site directory clean because this directory always tends to get full by 

- Old index.html files (backup)
- Media / static content directories
- Notes to editors
- Direct content which you put there to share it just this moment [I know ... this is bad practise] but forget to clean up
- Featured sub-sites

In order to manage all these files (and not to loose overview on clean-ups thus making dangerous deletions) it is IMHO important to stash helper files which don't have to do with this particular website.

And you can more easily prevent access by .htaccess e. g.

And a clean web directory (with two CMS files instead of 13) has less puzzling impact on the editors / administrators.

PhrozenByte:
Thank you for your contribution @jjtmp! :+1: However, as explained above, I think there's a better solution to achieve what you're trying to do. Thus I will close this now. If you have something to add, please don't hesitate to comment. @ALL, more feedback is still appreciated!

Issue #509, opened 2019-09-10
mbstring error on Debian Buster.


kc0dxf:
I'm migrating to Debian Buster.  I have php version 7.3 and also the php-mbstring package installed.  When I try to go to a fresh pico install I get this error: 
Pico requires the PHP extension 'mbstring' to run

What must I do to get it to work?
Thanks!

kc0dxf:
Thanks for the help.  What happened is that I upgraded Debian Stretch to Debian Buster.  Stretch must have had php 7.0 and when I upgraded to Buster, php 7.3 was installed.  However, Apache was still using php 7.0.  I simply did the following:
a2dismod php7.0
a2enmod php7.3
systemctl restart apache2

Now Pico is working fine!  Thanks for the suggestions.

JamesTheAwesomeDude:
I had this error as well (on the UAH webpages server, which I, obviously do not have administrative access to, thus could not [fix](https://stackoverflow.com/q/8266396) this at a system level)

My solution, which I'm noting here for posterity, was to include an [mbstring polyfill](https://github.com/symfony/polyfill-mbstring)

or, in a nutshell:

	php ~/composer.phar create-project picocms-pico-composer ./blog
	php ~/composer.phar require symfony/polyfill-mbstring -d ./blog

I have a fork created to include this require OOTB; I'm not sure how broad this case is. I'm sure there's lots of people on budget webhosts who'd love/appreciate such an inclusion. (Also, apparently some web frameworks explicitly require said administrative changes NOT to be set; this polyfill at a directory-level allows Pico to coexist on the same server as such frameworks.)

PR #517, opened 2019-11-08
Enhancement: require mbstring polyfill

JamesTheAwesomeDude:
c/f [#509/551912485](https://github.com/picocms/Pico/issues/509#issuecomment-551912485)

On some webhosts, `php-mbstring` may not be available; including [this 27KiB library](https://github.com/symfony/polyfill-mbstring) remedies those cases.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "symfony/yaml" : "^2.8",
         "erusev/parsedown": "^1.8|1.8.0@beta",
         "erusev/parsedown-extra": "^0.8|0.8.0@beta",
-        "symphony/polyfill-mbstring": "^1.12"
+        "symfony/polyfill-mbstring": "^1.12"
     },
     "suggest": {
         "picocms/pico-theme": "Pico requires a theme to actually display the contents of your website. This is Pico's official default theme.",
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,8 @@
         "twig/twig": "^1.35",
         "symfony/yaml" : "^2.8",
         "erusev/parsedown": "^1.8|1.8.0@beta",
-        "erusev/parsedown-extra": "^0.8|0.8.0@beta"
+        "erusev/parsedown-extra": "^0.8|0.8.0@beta",
+        "symphony/polyfill-mbstring": "^1.12"
     },
     "suggest": {
         "picocms/pico-theme": "Pico requires a theme to actually display the contents of your website. This is Pico's official default theme.",


PhrozenByte:
I remember that I've had this discussion before, but for some reason I can't find it anymore :unamused: There are two issues with `symfony/polyfill-mbstring` in my opinion:

1. `symfony/polyfill-mbstring` requires PHP's `iconv` extension. Even though `iconv` is a default extension, there definitely are installations out there which don't have `iconv` enabled. There's also a polyfill for `iconv`, however, `symfony/polyfill-iconv` is a pure PHP implementation, making it perform poorly; `symfony/polyfill-mbstring` is a partially native implementation at least. So, just to make this clear, using `symfony/polyfill-mbstring` is an improvement without any doubt. But it's no bullet-proof solution. Anyway, this isn't really a impediment.

2. `symfony/polyfill-mbstring` isn't really a recommended solution, rather an "last resort" type of solution. Even though it's a partially native implementation, the downsides are still notable. It's always better to install the native PHP extension - and for most webservers you'll have to do this anyway, because requiring `mbstring` is very common. Thus I'm not sure whether there are actually (budget) hosters out there which don't have `mbstring` installed (the number of support tickets about this topic would be awful I guess). I'd rather assume that this is mostly an issue in self hosting environments (or in situations like "the IT department gave me access to this server", i.e. neither "real" hosting, nor self-hosting). The better solution always is to install the native extension. Requiring `symfony/polyfill-mbstring` by default just "hides" this issue. Requiring it on a case-by-case basis is still possible though.

This is no decision, I'm still struggling about it... Since we're currently in beta for Pico 2.1 it's definitely the right time to make this decision. **More opinions would be appreciated!**

JamesTheAwesomeDude:
That definitely makes sense.

Seeing your concerns, I think I agree that (unconditionally) adding extra libraries that displace native, commonly-available functions with polyfills (however well-written), is probably not the right workaround.

I'm not sure how prevalent `mbstring` not being administratively available is; I know that https://webpages.uah.edu/ seems to lack it; but, other than that, I haven't really shopped around very much.

Perhaps it'd be a more elegant solution to just beef up the dependency documentation a bit? The biggest ~~annoyance~~ hurdle I had with this, was that nowhere in the [User Documentation](http://picocms.org/docs/) does it list any system requirements/dependencies beyond minimum PHP version.

If these were at least *noted somewhere*, I think it would enable the user to do appropriate troubleshooting without imposing the need for as much direct, technical debugging.

So, dependency-wise, what's missing from this list? If it's not much more than the following, would it be appropriate to just edit some extra deets into [`install.md`](https://github.com/picocms/picocms.github.io/blob/master/_docs/install.md)?

 * ~~PHP ≥ 5.3.6~~
 * `mod_rewrite`
 * `mbstring`

PhrozenByte:
Yeah, adding some note to the install instructions might be sufficient. Pico itself actually requires just PHP ≥ 5.3.6, however, Parsedown additionally requires the PHP extension `mbstring` and Parsedown Extra furthermore requires the PHP extension `dom`. That's it. `mod_rewrite` is, like URL rewriting in general, optional (unless you wanna use plugins implementing security stuff like authentication).

PR #535, opened 2020-03-29
Pico 3.0

PhrozenByte:
Since the issues with our old dependencies kinda got out of hand lately (see #528, #534 and 8a44584291685e5e163765ed6578c184c991794e), it's time to kick-start development of **Pico 3.0** :tada:

Major changes are going to be updated dependencies (Twig 3.0+ and Symfony YAML 5.0+) and lifting the requirements to PHP 7.2.5+. We definitely won't implement all of #317, but rather focus on lazy page discovery and loading. We definitely won't overhaul Pico's event system now, neither will we implement caching or i18n support with Pico 3.0 (maybe with Pico 3.1? :thinking:).

We'll also think about changing Pico's Markdown parser. Unfortunately Parsedown's development process got very unreliable lately. However, there's a lot to consider - not just BC, but also performance-wise (Parsedown is by far the most performant Markdown parser I know). Thus there's no decision made yet. Other opinions are very welcome on this topic! Also see e.g. #576

Pico 3.0 is a major release, thus it will break BC. As always, Pico's official [`PicoDeprecated` plugin](https://github.com/picocms/pico-deprecated) ([Pull Request](https://github.com/picocms/pico-deprecated/pull/2)) will do its best to maintain BC. However, be warned, we'll explicitly drop all pre-v1.0 and v1.0 behaviour that affects performance.

As always, **feedback is highly appreciated!** :heart: 

---

**Ideas**

- #581: Pico bundle package with various official plugins

---

**Release of Pico `v3.0.0-alpha.1`**

Since the mentioned issues are affecting live installations already, I decided to release a intermediate [Pico `v3.0.0-alpha.1`](https://github.com/picocms/Pico/releases/tag/v3.0.0-alpha.1). This release upgrades Pico's dependencies to Twig 2.12, Symfony YAML 3.4, Parsedown 1.7.4 and Parsedown Extra 0.8.1 for now (requiring PHP 7.0.8+). Both are still officially supported and are basically drop-in replacements, since they maintain BC on a best-effort basis ("best-effort" means "should work for most users", there can still be issues for some users). Pico `v3.0.0-alpha.1` is meant as a fast solution for all people experiencing the issues mentioned above.

**Feedback about Pico `v3.0.0-alpha.1` is highly appreciated**, especially regarding issues! :+1: 

---

**Related links**

- [List of ideas for Pico 3.0 and beyond](https://github.com/picocms/Pico/issues/317)
[...]

PR #585, opened 2021-05-17
Pico For 🚀 php >=8.x?

ILDaviz:
Hi pico team! 🚀 I'm update core pico for php> = 8 but not tested it's very very very early project ;D 
In this month i think complete it. And PicoLovers ❤️ Thanks for creating this piece of software!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "source": "https://github.com/picocms/Pico"
     },
     "require": {
-        "php": ">=7.0.8",
+        "php": ">=8.0",
         "ext-mbstring": "*",
         "twig/twig": "^2.12",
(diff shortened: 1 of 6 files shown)


PhrozenByte:
Thank you for your PR :heart:

Did you check out #535? A lot of this has been done already. Pico always was a project that aims for a maximum of compatibility, thus I'm not so sure whether lifting the minimum requirement to PHP 8.0 is suitable. This is not about whether PHP 8.0 is wide spread enough, it is, but there are still a lot of servers with older PHP versions, like Debian Buster servers with PHP 7.3, supported until 2024, or even worse, RHEL.

ILDaviz:
Ok @PhrozenByte I will then work on that version; D

PR #599, opened 2021-08-15
Update web links in index.md

dipohl:
Checked all web links and updated them where necessary especially eliminating some 404 errors.


PhrozenByte:
Thank you very much! :heart:

mayamcdougall:
@PhrozenByte Whoops. Didn't think to check the sample-content when removing the themes wiki. 😅

Don't let me forget to check through it when we eventually get the plugins migrated too!

PR #643, opened 2022-06-24
Replace strftime (obsolete in PHP 8.1)  by date

Marcussacapuces91:
I didn't dig very deep to see side effects


PhrozenByte:
Thank you for your contribution :+1: However, `date()` is no replacement for `strftime()`, as the latter supports different locales.

PR #680, opened 2023-12-01
Add Event Triggers to `content` Filter

PxaMMaxP:
In this merge I add important event triggers to improve the consistency in the functionality of PicoCMS. Previously, the `onContentPrepared` and `onContentParsed` events were only triggered when a page was called directly. 

This limitation led to problems with plugins that rely on these events, especially if the content was processed via the internal Twig filter `content`. The added event triggers in the `content` filter ensure that all plugins involved in content creation work smoothly even when the `content` filter is used. 

This change ensures broader and more reliable plugin compatibility within PicoCMS.

### DIFF
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -2169,7 +2169,9 @@ class Pico
                         $pageData = &$pages[$page];
                         if (!isset($pageData['content'])) {
                             $markdown = $pico->prepareFileContent($pageData['raw_content'], $pageData['meta'], $page);
+                            $this->triggerEvent('onContentPrepared', [ &$markdown ]);
                             $pageData['content'] = $pico->parseFileContent($markdown);
+                            $this->triggerEvent('onContentParsed', [ &$pageData['content'] ]);
                         }
                         return $pageData['content'];
                     }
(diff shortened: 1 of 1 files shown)


PhrozenByte (changes requested):
Thank you for your PR! :+1:

I assume you accidentally selected the wrong branch to merge, you probably rather meant to select your `content-filter-trigger-events` branch instead of your `pico-3.0` branch. You might wanna open another PR.

Besides, I'm afraid that we can't merge it the way it is, because the `onContentPrepared` and `onContentParsed` events were designed to only fire for the requested page, not for other pages as well. This will likely break a lot of plugins, they just don't expect that. For other pages we have the `onSinglePageLoading`, `onSinglePageContent`, and `onSinglePageLoaded` events. Since we indeed don't have events that fire before and after Markdown processing, we could indeed add such new events, just with a different name.

We'd have to let `PicoDeprecated` know about these new events as well.

PxaMMaxP:
Heyy,

thanks for your detailed explanations, and sorry for the confusion with the wrong branch.

About the change of the `content-filter-trigger-events` branch:
My problem was that when I rendered the content of several pages with the `content`-filter on one page, the corresponding plugins were not applied to that content.

In my specific case, it was the targetBlank plugin. Links were not changed accordingly on the content-->`content`-filter-->output path.

I noticed that there was no note (issue..) about this and it should have made me suspicious.

I'll close the pull request and take another close look at the whole thing.

PR #681, opened 2023-12-01
Enhanced Sorting by Nested Meta Values in Pages

PxaMMaxP:
This pull request introduces an enhancement to the sorting functionality within pages. Previously, sorting was limited to top-level meta values. With this update, the sorting logic now supports nested meta values, providing greater flexibility and precision in organizing content.

Key Changes:

- Extended the pages_order_by_meta configuration to interpret dot-separated strings as paths to nested meta values.
- Updated the sorting algorithm to recursively access nested meta values based on the provided path.
- 
This enhancement ensures a more dynamic and versatile approach to content organization, particularly beneficial for complex page structures with deeply nested meta data.

### DIFF
--- a/config/config.yml.template
+++ b/config/config.yml.template
@@ -30,8 +30,8 @@ twig_config:                        # Twig template engine config
 #
 date_format: "%D %T"                # Pico's default date format;
                                     #     See https://php.net/manual/en/function.strftime.php for more info
-pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta"); use '.' for nested meta keys (e.g., 'author.info')
-pages_order_by: alpha               # Change how Pico sorts pages ("alpha" for alphabetical order, "date", or "meta")
+pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta")
+                                    #     Use '.' notation for nested meta keys (e.g. 'author.info')
 pages_order: asc                    # Sort pages in ascending ("asc") or descending ("desc") order
 content_dir: ~                      # The path to Pico's content directory
 content_ext: .md                    # The file extension of your Markdown files
--- a/config/config.yml.template
+++ b/config/config.yml.template
@@ -30,7 +30,7 @@ twig_config:                        # Twig template engine config
 #
 date_format: "%D %T"                # Pico's default date format;
                                     #     See https://php.net/manual/en/function.strftime.php for more info
-pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta")
+pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta"); use '.' for nested meta keys (e.g., 'author.info')
 pages_order_by: alpha               # Change how Pico sorts pages ("alpha" for alphabetical order, "date", or "meta")
 pages_order: asc                    # Sort pages in ascending ("asc") or descending ("desc") order
 content_dir: ~                      # The path to Pico's content directory
--- a/config/config.yml.template
+++ b/config/config.yml.template
@@ -32,6 +32,7 @@ date_format: "%D %T"                # Pico's default date format;
                                     #     See https://php.net/manual/en/function.strftime.php for more info
 pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta")
                                     #     Use '.' notation for nested meta keys (e.g. 'author.info')
+pages_order_by: alpha               # Change how Pico sorts pages ("alpha" for alphabetical order, "date", or "meta")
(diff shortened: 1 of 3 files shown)


lib/Pico.php
```diff
@@ -1830,11 +1830,22 @@ protected function sortPages()
...
+            uasort($this->pages, function ($a, $b) use ($alphaSortClosure, $order, $orderByMetaKeys) {
+                $aMeta = $a['meta'];
+                $bMeta = $b['meta'];
+                // Iterate through the meta keys to drill down into nested arrays for the final sort value.
+                foreach ($orderByMetaKeys as $key) {
+                    $aMeta = isset($aMeta[$key]) ? $aMeta[$key] : null;
+                    $bMeta = isset($bMeta[$key]) ? $bMeta[$key] : null;
+                }
+
+                $aSortValue = $aMeta;
```
PhrozenByte, line 1845: Can't we use a single variable here?

lib/Pico.php
```diff
@@ -1830,11 +1830,22 @@ protected function sortPages()
         if ($orderBy === 'meta') {
             // sort by arbitrary meta value
             $orderByMeta = $this->getConfig('pages_order_by_meta');
-            uasort($this->pages, function ($a, $b) use ($alphaSortClosure, $order, $orderByMeta) {
-                $aSortValue = isset($a['meta'][$orderByMeta]) ? $a['meta'][$orderByMeta] : null;
+            // Split the configuration value into an array of keys to allow sorting by nested meta values.
```
PhrozenByte, line 1833: The code is pretty self-explanatory, I don't think that we need inline code comments here (same with the comment below). However, we do need documentation in `config.yml.template` and `CHANGELOG.md`.

PhrozenByte (changes requested):
Thank you for your PR! :+1:

LGTM besides some minor things.

Since it's fully backwards compatible we don't need any changes in `PicoDeprecated`.

lib/Pico.php
PxaMMaxP: I had to think a bit about what you mean exactly. But it should be simpler and still quite clear.

Thank you for the opportunity to contribute here! It's my first time contributing publicly.

PxaMMaxP:
I would like to extend my apologies for the recent commit and subsequent revert commit. Due to the challenges encountered, I am providing the suggested line for the changelog here. This is to ensure clarity and avoid any further complications.

Suggested Changelog Entry:
```
* [New] Enhance `pages_order_by_meta` functionality to allow sorting by nested meta values using '.' notation (e.g., 'author.info')
```

If this approach poses any issues, I am more than willing to open an additional pull request to handle this change. Please let me know if that would be preferable.

Thank you for your understanding and support.

PxaMMaxP:
Heyy,

thanks for your detailed explanations, and sorry for the confusion with the wrong branch.

About the change of the `content-filter-trigger-events` branch:
My problem was that when I rendered the content of several pages with the `content`-filter on one page, the corresponding plugins were not applied to that content.

In my specific case, it was the targetBlank plugin. Links were not changed accordingly on the content-->`content`-filter-->output path.

I noticed that there was no note (issue..) about this and it should have made me suspicious.

I'll close the pull request and take another close look at the whole thing.

config/config.yml.template
```diff
@@ -30,7 +30,7 @@ twig_config:                        # Twig template engine config
 #
 date_format: %D %T                  # Pico's default date format;
                                     #     See https://php.net/manual/en/function.strftime.php for more info
-pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta")
+pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta"); use '.' for nested meta keys (e.g., 'author.info')
```
PhrozenByte, line 33:
```suggestion
pages_order_by_meta: author         # Sort pages by meta value "author" (set "pages_order_by" to "meta")
                                    #     Use '.' notation for nested meta keys (e.g. 'author.info')
```

PhrozenByte (changes requested):
LGTM :+1:

See the minor suggested change below. Your suggested changelog addition looks good, to. Can you please add it to this PR? Will merge it then. Thanks!

PxaMMaxP:
A little tinkering led to success: The changelog has now been changed accordingly!

Thanks again for your patience!

PhrozenByte:
Looks like there went something wrong, the line wrapping in `config.yml.template` was lost again (I assume because you didn't pull after you've committed the change via GitHub's "Commit suggestion" feature; you can use `git pull --rebase` if you have unpushed commits locally). While you're at it, can you also add line wraps in `CHANGELOG.md` matching the line length of the other entries as well? Thanks again! :+1:

PhrozenByte (approved):
LGTM, great work, thank you very much! :heart: :tada: :+1:

Don't worry about the failing CI btw... :wink:

PR #683, opened 2023-12-04
Add meta data attribute LANG to support multi language websites

dipohl:
I added this for my multi language blog. I think it can be useful also for other users and will be glad when you integrate the addition.

### DIFF
--- a/lib/Pico.php
+++ b/lib/Pico.php
@@ -1450,6 +1450,7 @@ class Pico
                 'Robots' => 'robots',
                 'Template' => 'template',
                 'Hidden' => 'hidden',
+                'Lang' => 'lang',
             ];
 
             if ($this->themeMetaHeaders) {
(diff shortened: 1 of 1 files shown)


PhrozenByte:
Thank you for your contribution! :+1:

The pre-registered meta headers exist for backwards compatibility reasons only, since Pico 2.1 you can use [`pico-theme.yml` (see default theme)](https://github.com/picocms/pico-theme/blob/master/pico-theme.yml) to register additional meta headers. The only difference is that you can't access it via `{{ page.lang }}`, but `{{ page.meta.lang }}` (just for the record, `{{ page.title }}` also exists for backwards compatibility reasons only, one should use `{{ page.meta.title }}` instead).

dipohl:
ah, ok. Thanks for the hint and explanation. I will have a look at the new method then. I appreciate that you implemented a more flexible one! :+1:

